PR #314, opened 2023-09-14
Basic export in Mermaid format

kcaswick:
This exports in Mermaid flowchart format, with all the same features as the D2 export, but does not attempt to export links or other enhancements available in Mermaid.


davydkov:
Wow, thanks @kcaswick 

Also, your PR discloses I have not configured GitHub actions correctly for external PRs

PR #899, opened 2024-08-13
Replaced unix path separator with path.sep to enable build on windows

pavelpykhtin:
Fixes build error "../icons/all.js (3:15): Expected unicode escape" when building from git bash


davydkov (approved):

davydkov:
Thank you @pavelpykhtin

PR #930, opened 2024-08-23
Added metadata field which could be used to add custom attributes to the model

pavelpykhtin:
Metadata field is used to annotate elements and relationships with custom attributes which later could be used in other artifacts built from architecture model.


davydkov:
Nice job @pavelpykhtin, I like it!

Can you please add some smoke tests, like any from [here](https://github.com/likec4/likec4/tree/main/packages/language-server/src/__tests__)

pavelpykhtin:
> Can you please add some smoke tests, like any from [here](https://github.com/likec4/likec4/tree/main/packages/language-server/src/__tests__)

Sure

davydkov (approved):

PR #945, opened 2024-08-28
fix(dsl): ambiguous alternatives

davydkov:
[log] [lsp] Ambiguous Alternatives Detected: <0, 1> in <OPTION1> inside <RelationFragment​> Rule, <String, }, }> may appears as a prefix path in all these alternatives. See: https://chevrotain.io/docs/guide/resolving_grammar_errors.html#AMBIGUOUS_ALTERNATIVES For Further details.
[log] [lsp] Ambiguous Alternatives Detected: <0, 1> in <OPTION1> inside <RelationFragment​> Rule, <String, IdTerminal, ->> may appears as a prefix path in all these alternatives. See: https://chevrotain.io/docs/guide/resolving_grammar_errors.html#AMBIGUOUS_ALTERNATIVES For Further details.


davydkov:
Fixed in #947

Issue #916, opened 2024-08-22
[Feature request] Custom Colors


davydkov:
### Discussed in https://github.com/likec4/likec4/discussions/855

<div type='discussions-op-text'>

<sup>Originally posted by **simonwfarrow** July 19, 2024</sup>
Hi, I was wondering if it would be possible to allow custom colors for a style by supplying a hex value?</div>

PR #958, opened 2024-09-01
Implemented support for custom colors (#916)

pavelpykhtin:
Custom colors could be defined in spec with given name and hex value.
The defined color then could be used as one of a theme color.

Stylesheet with cssvars for custom colors is rendered in diagram component. Thees styles uses same ':where[data-likec4-color=<color-name>]' selector as theme styles.
Also tried [alternative](https://github.com/pavelpykhtin/likec4/tree/feat/custom-colors) with inline definition (use hex value instead of color name) but there are some problems with passing overridden cssvars into m.div component. Moreover, such implementation seems to be pretty messy.

### DIFF
--- a/packages/core/src/colors/index.ts
+++ b/packages/core/src/colors/index.ts
@@ -1,11 +1,7 @@
-import type {
-  HexColorLiteral,
-  LikeC4Theme,
-  ThemeColorValues,
-} from '../types/theme'
+import { generateColors } from '@mantine/colors-generator'
+import type { HexColorLiteral, LikeC4Theme, ThemeColorValues } from '../types/theme'
 import { ElementColors } from './element'
 import { RelationshipColors } from './relationships'
-import { generate } from '@ant-design/colors'
 
 export const defaultTheme = {
   elements: ElementColors,
@@ -15,19 +11,26 @@ export const defaultTheme = {
 } satisfies LikeC4Theme
 
 export function computeColorValues(color: HexColorLiteral): ThemeColorValues {
-  const colors = generate(color)
+  if (color.match(/^#([0-9a-f]{3}){1,2}$/i)) {
+    const colors = generateColors(color)
 
-  return {
-    elements: {
-      fill: color,
-      stroke: colors[6] as HexColorLiteral,
-      hiContrast: colors[1] as HexColorLiteral,
-      loContrast: colors[2] as HexColorLiteral
-    },
-    relationships: {
-      lineColor: color,
-      labelColor: colors[4] as HexColorLiteral,
-      labelBgColor: colors[9] as HexColorLiteral
+    return {
+      elements: {
+        fill: colors[6] as HexColorLiteral,
+        stroke: colors[7] as HexColorLiteral,
+        hiContrast: colors[0] as HexColorLiteral,
+        loContrast: colors[1] as HexColorLiteral
+      },
+      relationships: {
+        lineColor: colors[4] as HexColorLiteral,
+        labelColor: colors[3] as HexColorLiteral,
+        labelBgColor: colors[9] as HexColorLiteral
+      }
+    }
+  } else {
+    return {
+      elements: defaultTheme.elements['primary'],
+      relationships: defaultTheme.relationships['primary']
     }
   }
 }
--- a/packages/core/src/colors/index.ts
+++ b/packages/core/src/colors/index.ts
@@ -1,6 +1,11 @@
-import type { LikeC4Theme } from '../types/theme'
+import type {
+  HexColorLiteral,
+  LikeC4Theme,
+  ThemeColorValues,
+} from '../types/theme'
 import { ElementColors } from './element'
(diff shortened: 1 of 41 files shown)


packages/diagram/src/LikeC4Diagram.tsx
```diff
@@ -70,6 +71,7 @@ export function LikeC4Diagram({
           fitView={fitView}
           {...initialRef.current}
         >
+          <LikeC4CustomColors customColors={view.customColorDefinitions}></LikeC4CustomColors>
```
davydkov: Make sense wrap with `if`, if there is any customColorDefinition

packages/diagram/src/LikeC4CustomColors.tsx
```diff
@@ -0,0 +1,34 @@
...
+        return ref.substring(4, end == -1 ? ref.length - 1 : end);
+    }
+
+    const styles = entries(customColors)
+        .map(([name, color]) => toStyle(name, color))
+        .join('\n');
+
+    return <>
+        <style>
+            {styles}
```
davydkov: I am not sure if React may escape characters, and I did via `dangerouslySetInnerHTML`

https://github.com/likec4/likec4/blob/b0646a37e4b01555c63b22f1c28ec3b9a4e2d089/packages/likec4/app/react/components/LikeC4ViewElement.tsx#L105-L115

packages/core/src/colors/index.ts
```diff
@@ -1,6 +1,11 @@
-import type { LikeC4Theme } from '../types/theme'
+import type {
+  HexColorLiteral,
+  LikeC4Theme,
+  ThemeColorValues,
+} from '../types/theme'
 import { ElementColors } from './element'
 import { RelationshipColors } from './relationships'
+import { generate } from '@ant-design/colors'
```
davydkov: I suggest to use [`@mantine/colors-generator`](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/colors-generator/src/generate-colors.ts#L44C17-L44C31)

This is how it works: https://mantine.dev/colors-generator/

Why?
Initially I used tailwind, and picked colors manually, thats why I have these constants here. I am not good in knowing the aspects, and seems picking something like [Open Colors](https://yeun.github.io/open-color/), [Radix](https://www.radix-ui.com/colors) or [Adobe Spectrum](https://spectrum.adobe.com/page/color-system/)

At least for now I suggest to stay with mantine scale.

packages/core/src/colors/index.ts
```diff
@@ -9,4 +14,22 @@ export const defaultTheme = {
...
+export function computeColorValues(color: HexColorLiteral): ThemeColorValues {
+  const colors = generate(color)
+
+  return {
+    elements: {
+      fill: color,
+      stroke: colors[6] as HexColorLiteral,
+      hiContrast: colors[1] as HexColorLiteral,
+      loContrast: colors[2] as HexColorLiteral
+    },
```
davydkov: If we use [@mantine/colors-generator](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/colors-generator/src/generate-colors.ts#L44C17-L44C31)

```suggestion
    elements: {
      fill: color[6] as HexColorLiteral,
      stroke: colors[7] as HexColorLiteral,
      hiContrast: colors[0] as HexColorLiteral,
      loContrast: colors[1] as HexColorLiteral
    },
```

packages/diagram/package.json
```diff
@@ -45,6 +45,7 @@
     "clean": "run -T rimraf -v -g 'dist/*' .tsbuildinfo"
   },
   "dependencies": {
+    "@ant-design/colors": "^7.1.0",
```
davydkov: Should be moved to `core` where it is used actually

davydkov (changes requested):
You gave it a nice shot!

Though changes in langium indeed seem hacky.

I am working on [feat/model-api](https://github.com/likec4/likec4/tree/feat/model-api) branch (that exposes specifications in model) and 
I am afraid we may have lot merge conflicts 
Can you try to rebase?

pavelpykhtin:
Hi, @davydkov ,

Thank you for feedback! I'll apply fixes in a day or two.

packages/diagram/package.json
pavelpykhtin: Done

pavelpykhtin:
@davydkov ,

Fixed issues from review and rebased on top of feat/model-api.

davydkov (approved):
Would be nice:
- add validation that custom color is not one of theme colors
- move `colorDefinitions` from views to specifications https://github.com/likec4/likec4/blob/4b5f6c183898a69d7699bf96ec9633dcd4ad9c43/packages/core/src/types/model.ts#L10-L14
- views include only relevant `colorDefinitions` (after compute)

davydkov:
Sorry, I was busy last days
Thank you for this PR
I fix merge issues

Resolves #916

PR #979, opened 2024-09-11
Implementation of formatter for likec4 documents

pavelpykhtin:
Hey @davydkov 

I've implemented a simple vscode formatter with langium. It applies basic cleanup rules (adds indentation, removes/adds spaces where needed etc.).

Would you mind including it into the existing extension?


davydkov (commented):
PR is awesome!!!

I'm curious, is it possible to keep line breaks and empty lines?
Let's say this example:

```zig
model {
  el1 = element 'El1'

  el2 = element 'El2' {
   description ''
  
   el21 = element '21'
   el22 = element '22'

   el21 -> el22 #tag1
  }
}

views {
  view index {
    title '...'

    include
      el1,
      el2._

    exclude * -> el2

    include
      * -> *
      where
        tag is #tag1
      with {
        color red 
      }
  }

  view index2 {
  }
}
```

Turns to

```zig
model {
  el1 = element 'El1'
  el2 = element 'El2' {
    description ''
    el21 = element '21'
    el22 = element '22'
    el21 -> el22 #tag1
  }
}

views {
  view index {
    title '...'
    include
      el1,
      el2._
    exclude * -> el2
    include
      * -> *
      where tag is #tag1 with {
        color red
      }
  }
  view index2 {
  }
}
```

Can we try to keep initial formatting as close as possible?

And would it better to use [`toMatchInlineSnapshot`](https://vitest.dev/api/expect.html#tomatchinlinesnapshot) in tests?

pavelpykhtin:
It's easy to keep empty lines between expressions (done).
But when it comes to line breaks inside the expression (like in the 'include' expression in the example above) things become more complicated. It definitely could be done, and I'm planing further improvements, but I don't know how much time it will take.

Thanks for pointing to inline snapshots!

Issue #1002, opened 2024-09-13
VS Code visual editor problem


anatoly-kryzhanovsky:
i faced strange behavior of vs code editor:
- i move some relation ship (delete curve point, move other)
- after few second i try to move another relationship but after i stop drag it return to initial state

i see that layout scheme in my c4 file was updated and there is enought delay between operation for data writen to disk

![demo](https://github.com/user-attachments/assets/e5645a1f-35ba-436c-ae11-6e146e684bb7)

version of plugin 1.10.0

PR #1008, opened 2024-09-16
Fix edge changes being reverted on mouse up (#1002)

pavelpykhtin:
It seems that the problem described in #1002 (at least one) is caused by the xyflow state being reset to initial state on mouse up.
My guess is that xyflow state is overwritten by diagram state on edge selection, which is part of the xyflow component.

As a short term solution, I've added selection of the edge before any changes are made.

As a long term solution, it seems that some changes in diagram state management should be made. As far as I can understand, the actual cause of the problem is that there are two versions of the state: one is the current state of the xyflow component (used for rendering), the other is the previous state (stored in file? or in a some intermediate object?). Thus, those two versions are must be merged into one or the sync procedure must be improved. Unfortunately, I don't have deep enough understanding of the current implementation and xyflow library particularly to implement this. @davydkov , may be you have some ideas on this?

### DIFF
--- a/packages/diagram/src/xyflow/edges/RelationshipEdge.tsx
+++ b/packages/diagram/src/xyflow/edges/RelationshipEdge.tsx
@@ -307,7 +307,12 @@ export const RelationshipEdge = /* @__PURE__ */ memo<EdgeProps<XYFlowEdge>>(func
     if (!domNode || e.pointerType !== 'mouse') {
       return
     }
+
     const { xyflow } = diagramStore.getState()
+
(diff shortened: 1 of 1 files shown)


davydkov:
> Unfortunately, I don't have deep enough understanding of the current implementation 

Yes, me neither 😄 

Edge editing was super experimental.
[This repository](https://github.com/idootop/reactflow-auto-layout?tab=readme-ov-file#2-edge-auto-routing) seems really nice, but I had no chances yet to try it.

Thanks for contribution!

PR #1013, opened 2024-09-22
Minor improvements to edge editing

pavelpykhtin:
Fixed issue with control point deletion being reset on selection change
Added control point insertion on right click
Added "Reset all control points" command to simplify layout editing. **Before**:
before
After:
after

### DIFF
--- a/packages/diagram/src/ui/top-left/TopLeftPanel.tsx
+++ b/packages/diagram/src/ui/top-left/TopLeftPanel.tsx
@@ -16,7 +16,8 @@ import {
   IconClearAll,
   IconFileSymlink,
   IconFocusCentered,
-  IconMenu2
+  IconMenu2,
+  IconRouteOff
 } from '@tabler/icons-react'
 import clsx from 'clsx'
 import { AnimatePresence, m } from 'framer-motion'
@@ -147,7 +148,7 @@ const ResetControlPointsButton = () => {
             e.stopPropagation()
             store.getState().resetEdgeControlPoints()
           }}>
-          <IconClearAll />
+          <IconRouteOff />
         </ActionIcon>
       </m.div>
     </AnimatePresence>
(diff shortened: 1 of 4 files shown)


packages/diagram/src/ui/top-left/TopLeftPanel.tsx
```diff
@@ -127,6 +129,24 @@ const LayoutDriftNotification = (props: PopoverProps) => (
   </HoverCard>
 )
 
+const ResetControlPointsButton = () => {
+  const store = useDiagramStoreApi()
+
+  return (
+    <AnimatePresence>
```
davydkov: `<AnimatePresence>` is redundant here, as `ResetControlPointsButton` already inside one from the line 193

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
```diff
@@ -127,6 +129,24 @@ const LayoutDriftNotification = (props: PopoverProps) => (
...
+          <Tooltip label="Reset all control points">
```
davydkov: Better add  `{...portalProps}` to `Tooltip`

TLTR; when diagram is inside shadow root (generated react/webcomponent) - react portals also should be inside that root (see for other buttons)

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
```diff
@@ -183,6 +203,7 @@ export const TopLeftPanel = () => {
             <LayoutDriftNotification {...portalProps} />
           </m.div>
         )}
+        <ResetControlPointsButton />
```
davydkov: Let show this button, if only
`readonly !== true && experimentalEdgeEditing === true`

pavelpykhtin: Done

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
pavelpykhtin: Done, thank you for the explanation!

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
pavelpykhtin: Done

PR #1038, opened 2024-09-25
feat: style definition applicable to all views

hubertmis:
Add a feature to define a style applicable to all views. It is useful to define in a unified way how e.g. elements tagged as #deprecated are styled in all the diagrams.

Please note, I'm an embedded developer and I was learning about TypeScript on the go. I'm happy to fix any issues in my code and broaden my higher-level language horizons :)

This is intended to fix https://github.com/likec4/likec4/discussions/961

### DIFF
--- a/apps/docs/src/content/docs/dsl/views.mdx
+++ b/apps/docs/src/content/docs/dsl/views.mdx
@@ -523,11 +523,46 @@ view apiApp of internetBankingSystem.apiApplication {
   // apply to elements not tagged
   style element.tag != #deprecated {
     opacity 20%
-  }  
+  }
 }
 ```
 
 Please note, that [overrides](#with-overrides) have higher priority.
+
+Styles can be also applied to all views
+
+```likec4
+views {
+  // apply to all elements in all views
+  style * {
+    color muted
+    opacity 10%
+  }
+
+  // apply in all views to elements with specific tag
+  style element.tag = #deprecated {
+    color muted
+  }
+
+  // apply in all views to elements not tagged
+  style element.tag != #deprecated {
+    opacity 20%
+  }
+
+  view apiApp of internetBankingSystem.apiApplication {
+    include *
+
+    // override global style
+    style element.tag = #deprecated {
+      color red
+    }
+  }
+
+  view storeConnector of internetBankingSystem.storeConnector {
+    include *
+  }
+}
+```
 
 ### Extend views
 
--- a/packages/core/src/types/model.ts
+++ b/packages/core/src/types/model.ts
@@ -1,7 +1,7 @@
 import type { UnwrapTagged } from 'type-fest'
 import type { Element, ElementKind, ElementKindSpecification, Fqn, Tag } from './element'
 import type { Relation, RelationID, RelationshipKindSpecification } from './relation'
-import type { ComputedView, DiagramView, LikeC4View, ViewID } from './view'
+import type { ComputedView, DiagramView, LikeC4View, ViewID, ViewRule } from './view'
 
 /**
  * Parsed elements, relations, and views.
@@ -15,6 +15,7 @@ export interface ParsedLikeC4Model {
   elements: Record<Fqn, Element>
   relations: Record<RelationID, Relation>
   views: Record<ViewID, LikeC4View>
+  global_rules?: ViewRule[]
 }
 
 /**
--- a/packages/language-server/src/__tests__/views.spec.ts
+++ b/packages/language-server/src/__tests__/views.spec.ts
@@ -233,6 +233,21 @@ describe.concurrent('views', () => {
       }
     }`
 
+  test('global ViewStyleRules - valid').valid`${model}
+    views {
+      style * {
+        color: secondary
+      }
+      view {
+        include *
+        exclude -> frontend
+      }
+      style backend, infra {
+        color: muted
+      }
+    }
+    `
+
   test('trailing comma in predicates').valid`${model}
     views {
       view {
--- a/packages/language-server/src/ast.ts
+++ b/packages/language-server/src/ast.ts
@@ -131,6 +131,13 @@ export const ViewOps = {
   }
 }
 
+export interface ParsedAstGlobalRule {
+  __: 'global_rule'
+  rules: c4.ViewRule[]
+}
+
+export type ParsedAstViewOrRule = ParsedAstView | ParsedAstGlobalRule
+
 export interface ParsedLink {
   title?: string
   url: string
@@ -164,6 +171,7 @@ export interface LikeC4DocumentProps {
   c4Elements?: ParsedAstElement[]
   c4Relations?: ParsedAstRelation[]
   c4Views?: ParsedAstView[]
+  c4GlobalRules?: c4.ViewRule[]
   // Fqn -> Element
   c4fqnIndex?: MultiMap<c4.Fqn, DocFqnIndexAstNodeDescription>
 }
@@ -190,7 +198,8 @@ export function cleanParsedModel(doc: LikeC4LangiumDocument) {
     },
     c4Elements: [],
     c4Relations: [],
-    c4Views: []
+    c4Views: [],
+    c4GlobalRules: []
   }
   return Object.assign(doc, props) as ParsedLikeC4LangiumDocument
 }
@@ -213,6 +222,7 @@ export function isParsedLikeC4LangiumDocument(
     && !!doc.c4Elements
     && !!doc.c4Relations
     && !!doc.c4Views
+    && !!doc.c4GlobalRules
     && !!doc.c4fqnIndex
   )
 }
--- a/packages/language-server/src/like-c4.langium
+++ b/packages/language-server/src/like-c4.langium
@@ -199,8 +199,9 @@ ModelViews:
   '}';
 
 type LikeC4View = ElementView | DynamicView;
-LikeC4ViewRule returns LikeC4View:
-  ElementView | DynamicView;
+type LikeC4ViewOrRule = LikeC4View | ViewRule;
+LikeC4ViewRule returns LikeC4ViewOrRule:
+  ElementView | DynamicView | ViewRule;
 
 ElementView:
   'view' name=Id? (
--- a/packages/language-server/src/lsp/CodeLensProvider.ts
+++ b/packages/language-server/src/lsp/CodeLensProvider.ts
@@ -1,7 +1,7 @@
 import { DocumentState, type LangiumDocument, type MaybePromise } from 'langium'
 import type { CodeLensProvider } from 'langium/lsp'
 import type { CancellationToken, CodeLens, CodeLensParams } from 'vscode-languageserver'
-import { isParsedLikeC4LangiumDocument, ViewOps } from '../ast'
+import { ast, isParsedLikeC4LangiumDocument, ViewOps } from '../ast'
 import { logger } from '../logger'
 import type { LikeC4Services } from '../module'
 
@@ -23,7 +23,7 @@ export class LikeC4CodeLensProvider implements CodeLensProvider {
     if (!isParsedLikeC4LangiumDocument(doc)) {
       return
     }
-    const views = doc.parseResult.value.views.flatMap(v => v.views)
+    const views = doc.parseResult.value.views.flatMap(v => v.views).filter(v => ast.isLikeC4View(v))
     return views.flatMap<CodeLens>(ast => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const viewId = ViewOps.readId(ast)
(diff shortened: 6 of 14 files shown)


davydkov:
Hey @hubertmis, great job 👏🏻👏🏻 ! And quite elegant considering this:

SCR-20240925-ueap

Though I'm skeptical about making these global. Let me explain.

My first thought when looking at this syntax was that these rules are scoped to the current views block and are only applied to nested views.
While I wouldn't say this breaks the Open/Closed principle, imagine a scenario where someone in your organization creates another .c4 file with global styles that satisfy their needs but interfere with others.

I like the idea, but I would keep it scoped to the current views. 
I think the implementation would be simpler this way. Instead of introducing `c4GlobalRules`, we could just prepend these "global" rules to each parsed view directly in the model parser.
No need to change compute views (by the way, you missed dynamic views).

What do you think?

I'd also suggest adding:

- Parser smoke tests, see examples [views-2.spec.ts](https://github.com/likec4/likec4/blob/b9df8185d71e64a62a80e7d18bbb052e121d28fe/packages/language-server/src/__tests__/views-2.spec.ts)
- Tests in `model-builder.spec.ts` to ensure the styles are applied correctly

hubertmis:
> I like the idea, but I would keep it scoped to the current views.

Sounds good. It still meets my needs. I'll try to change the solution thr way you suggest.

> by the way, you missed dynamic views)

I did this on purpose to limit the scope of the changes for now.

> I'd also suggest adding:

I had a feeling the tests are missing, but I didn't have a clue where to look for them. Not to mention how to run them. Thanks for the hint!

davydkov:
> I had a feeling the tests are missing, but I didn't have a clue where to look for them. Not to mention how to run them. Thanks for the hint!

First ensure `yarn build` in root or default build task in VSCode.
And start `yarn vitest:ui` in root (run all tests) or inside `packages/language-server`.
You can keep it running - it reruns tests on file changes.

Changes in `*.langium` require code generation. This is done by either:
- Default build task in VSCode
- `yarn generate` in `packages/language-server`

Or better keep running `yarn watch:langium` in folder `packages/language-server`.

To debug tests - [vitest extension](https://marketplace.visualstudio.com/items?itemName=vitest.explorer) or check their docs

hubertmis:
> My first thought when looking at this syntax was that these rules are scoped to the current views block and are only applied to nested views.

After looking more into this solution, I don't think we can properly apply these "global rules" to the nested views without significant modifications to the syntax.

If I understand correctly, the parser iterates through the .likec4 files in the workspace folder in any order and the views are applied to `doc.c4Views` on the fly without any control over the hierarchy. So, if I created

```
views {
  style element.tag = #deprecated {
    color red
  }
}
```

in my index.likec4 containing the specification, I would expect this style to be applied to all the views being children of this file in my model hierarchy. But I don't see a way to express this hierarchy now.

If my understanding is correct, and I changed this feature as you suggested, I would need to copy
```
  style element.tag = #deprecated {
    color red
  }
```
part all over my files in each `views {}` block. I would prefer to avoid it.

If you see the proposed syntax confusing, what do you think about applying the global view rules in the `specification {}` block instead?

hubertmis:
In other words, I need a syntax that would allow me to define tag styles in one file that would be applicable to all views defined in all files. I'm open to suggestions on how to sketch this syntax to be aligned with the vision of likec4.

I was considering the following options:

1:
```
specification {
  ...
}

views {
  style element.tag = ... {
    ...
  }
}
```

2:
```
specification {
  style element.tag = ... {
    ...
  }
}
```

3:
```
specification {
  ...
}

styles {
  style element.tag = ... {
    ...
  }
}
```

@davydkov, what's your suggestion?

pavelpykhtin:
Hi, @hubertmis !

Good job! It's gone be very helpful feature.
In my opinion the second option is most suitable to express "global style" concept.

> But I don't see a way to express this hierarchy now.
Probably this could be implemented with 'extend view' feature: create an "abstract" base view with base styles and extend it in child views.

hubertmis:
> In my opinion the second option is most suitable to express "global style" concept.
> 

Should I then modify the PR to define global styles like this?

```
specification {
  style element.tag = ... {
    ...
  }
}
```

Is this agreed? Or should we discuss the options before I proceed?
I'm not sure how this repository is governed :)

> Probably this could be implemented with 'extend view' feature: create an "abstract" base view with base styles and extend it in child views.

I don't think that the extended views are enough to express the view hierarchy throughout the files. I see a challenge like this:

landscape.likec4:
```
specification {
  ...
}

model {
  environment system { ... }
}

views {
  // This is top-level views
  view of system {
    include *
  }
}
```

system.likec4:
```
model {
  extend system {
    component component1 {
    }
  }
}

views {
  // These are nested views
  // But the nesting can be only indirectly deducted based on the model
  view of component1 {
    ...
  }
}
```

I think that finding views hierarchy implicitly, based on the models hierarchy, and applying styles or other rules based on this implicit hierarchy would be a confusing syntax.

Alternatively, we could explicitly define the hierarchy by something like:

```
views {
  view of component1 {
    inherits_rules_from view.title = "system view"
  }
}
```

But then again the syntax gets complicated and I'm not sure if that gives real value.

I think most cases (at least all issues I need to solve) can be addressed by the global styles. I would stick to this feature, at least in this PR.

pavelpykhtin:
@hubertmis ,

> Should I then modify the PR to define global styles like this?
> Is this agreed? Or should we discuss the options before I proceed?
The final decision is on @davydkov. I'm here just as a future user of this feature :)

> I don't think that the extended views are enough to express the view hierarchy throughout the files.
I see. I meant explicit hierarchy of views (https://likec4.dev/dsl/views/#extend-views). It should work for non-scoped view, but in your example child view are scoped to component. It's not possible to extend non-scoped base view in scoped child view. Thus, it's not an option.

davydkov:
@hubertmis @pavelpykhtin

<details>
<summary>I vote for option 1</summary>

```zig
specification {
  ...
}

views {
  style element.tag = ... {
    ...
  }
}
```

</details>

However, I insist on making these styles "local" (I think _"local"_ describes it better than _"scoped"_).  
Consider this:

```zig
// Green views
views {
  style * {
    color green
  }

  view view1 {
  }
}

// Red views
views {
  style * {
    color red
  }

  view view2 {
  }
}
```

Pros:
- reduces boilerplate by sharing the same styles
- deterministic order and simple implementation
- looks very clear; almost no cognitive load in understanding

Cons:
- sharing is limited to the `views` block, not reusable.

I would say pros 2 and 3 are a good trade-off, and we still achieve some result, don't we?  
I suggest starting with this, and keeping all same-styled views in the same `views` block.

---

<details>
<summary><b>Here is a possible implementation</b></summary>

In Langium:  
https://github.com/likec4/likec4/blob/3201a2bf6b89e603ff000bd8b0c72f0d3e85ffd5/packages/language-server/src/like-c4.langium#L196-L199

I'm suggesting changing it to:

```langium
ModelViews:
  name='views' '{'
    (
      // Local styles
      styles+=ViewRuleStyle |
      views+=LikeC4ViewRule
    )*
  '}';
```

We separate views and local styles at the AST level.

In the model parser:  
https://github.com/likec4/likec4/blob/3201a2bf6b89e603ff000bd8b0c72f0d3e85ffd5/packages/language-server/src/model/model-parser.ts#L285-L299

First of all, we do not use `flatMap` on line 286 but rather iterate over each `views` block separately.

In each iteration, we first parse the `styles`.  
When we parse a view, we prepend local styles to its rules.

We don't even need to store these local styles somewhere; we prepend them during the parsing phase.

</details>

---

But I agree, the ultimate goal is sharing across the `views` block.  
And, as it seems to me, a good and, again, deterministic solution is option 3.

```zig
global styles {
  
  style1
    // reuse filter predicate
    where tag is #deprecated and ... {
      color red
    }  
}

views {  
  view {
    // Syntax Option 1, we need something to indicate "global":
    // like __.style1, ::style1, &style1
    include __.style1

    // Syntax Option 2 
    style.global style1 

    // Suggest yours
  }
}
```

Now, an interesting implication:

```zig
views {  

  // Apply global style to all views in a block  
  style.global style1

  view {
  }
}
```

---

I'm also considering the concept of _"namespacing"_ (or something like [domains](https://docs.icepanel.io/features/domains) in IcePanel). 

As you can see below, local styles also fit perfectly:

```zig
// By defining a path, we create a navigation tree and breadcrumbs
views "DataLake / Marketing / Ingestion" {

  // By default, all non-marketing elements are gray
  style element.tag != #marketing {
    color muted
  }

  // Marketing and ingestion elements are green
  style
    element.tag == #marketing,
    element.tag == #ingest {
      color green
    }

  dynamic view IngestPosthogToLake  {
    ...
  }

}
```

hubertmis:
@davydkov , thanks for the suggestions.

I would like to clarify one point:

> But I agree, the ultimate goal is sharing across the `views` block. And, as it seems to me, a good and, again, deterministic solution is suggested option 3.
> 
> ```zig
> global styles {
>   
>   style1
>     // reuse filter predicate
>     where tag is #deprecated and ... {
>       color red
>     }  
> }
> 
> views {  
>   view {
>     // Syntax Option 1, we need something to indicate "global"
>     // like __.style1, ::style1, &style1
>     include __.style1
> 
>     // Syntax Option 2 
>     style.global style1 
> 
>     // Suggest yours
>   }
> }
> ```

Assuming I create a few global styles that I would like to apply to all my views the code would look like:

```zig
global styles {
  
  style1
    // reuse filter predicate
    where tag is #deprecated and ... {
      color red
    }  

  style2 where tag is #unstable { ... }
  style3 where tag is #thirdparty { ... }
  style4 where tag is #contractor { ... }
}

views {  
  view {
    style.global style1
    style.global style2
    style.global style3
    style.global style4
  }
}
```

So I need to copy this `style.global ...` sequence through all my `view {}` nodes, or at least through all my `views {}` nodes (if local `views {}` styles get implemented). If a created `style5`, I would need to copy it across all files and nodes. This is a maintenance overhead I would like to avoid (one of the reasons I didn't decide to move to likec4 from structurizr yet :) ).

I would prefer to define a solution that applies a defined style to all matching elements, similar to the already existing specification element styles:
```zig
specification {
  element x {
    style {
    }
  }
}
```

I was considering making something similar. Another option would be:

```zig
specification {
  tag x {
    style {
    }
  }
}
```

Or even better, option 2 I suggested in the previous message because it could replace the specification element style, apply to the specification tag style, and potentially more like wildcards:

```zig
specification {
  element x
  tag y

  style element.kind = x { ... }
  style element.tag = #y { ... }
  style element.tag != #y { ... }
  style * { ... }
}
```

Pros:

1. Applicable to all views in the workspace - little maintenance when updating and creating global styles
2. Simple implementation (I already created a PoC on a [branch](https://github.com/likec4/likec4/commit/de878df0b2876750df51e06b30aadc3dab70da47))
3. Clear, simple syntax, similar to already existing
4. Can't be misused by adding global styling in sub-files, considering only one file can include `specification {}`.

Cons:

1. Missing granular control on applying only selected styles in `views {}` or `view {}` nodes

I believe the value of the granular styling control is not worth more complex syntax.

Moreover, the approach I sketched could be extended to solve another issue I described in #962 :

```zig
specification {
  exclude * where tag is #project1 and tag is not #project2
}
```

It allows an architect to define one specification file per project and share model files (including correctly tagged elements) between multiple projects.

hubertmis:
And another attempt: keeping option 2, but applying rules in the model parser. The specification type, model builder, and compute remain unchanged: https://github.com/likec4/likec4/commit/bad7ba4c7520e8072480cdc3c4423395f14de55d

hubertmis:
I started experimenting with global predicates to meet #962 requirements.

After writing a few tests, I like @davydkov 's proposal better, but I would suggest some enhancements to the syntax:

```zig
global {
  styles {
    // global styles are grouped in sets
    set style_set_1 {
      style where tag is ... {
        ...
      }
      style where tag is ... {
        ...
      }
    }

    style style_2 where tag is ... { // implicitly create a set 'style_2' containing 1 style
      ...
    }
  }

  predicates {
    set project1 {
      exclude * where tag is #project2 and tag is not #project1
    }
    predicate project2 exclude * where tag is #project1 and tag is not #project2
}

views {
  apply style style_set_1

  view {
    global style style_2
    include *
    global predicate project1
}
```

This seems to meet all @davydkov 's requirements, and introducing sets of rules meets my requirements of minimal maintenance when I modify the global styles. I expect to have a single set of styles in my diagrams that I apply to all views. I would also apply a few sets of predicates to most of my views (each view includes one of the global predicates' sets).

What do you think about this proposal?

davydkov:
@hubertmis Do we close this one?

hubertmis:
Replaced with #1058

PR #1044, opened 2024-09-26
Fix deleted control point being reverted on mouse up

pavelpykhtin:
A control point was deleted on the mouse down event. Thus, the mouse up event was routed to canvas. Canvas clears selected objects list on mouse events, which triggers edge update and reverts initial control points array.

Moved control point deletion into mouse up handler to ensure that mouse up event will be handled by control point and not by canvas.
Split onControlPointerDown into two handlers according to the pressed button.


davydkov:
👍🏻 👍🏻

Issue #1053, opened 2024-09-30
Add support for spacing parameters of autolayout


pavelpykhtin:
In some cases (especially for layered diagrams) default autolayout parameters does not work well:
![image](https://github.com/user-attachments/assets/d09d4101-eeaf-4df0-bbdb-fac7f43373ca)

In many cases, it is possible to solve the problem by increasing distance between nodes. Although it could be done with manual layout, autolayout is the better option because it does not require additional efforts to maintain.

Suggested syntax:
```
view v {
  autoLayout TopBottom 100 50
}
```

PR #1057, opened 2024-10-04
Added optional params to adjust autolayout (#1053)

pavelpykhtin:
New parameters allow adjusting distance between ranks and nodes.
Manual layout saved in older versions will be preserved.

Implements #1053


davydkov:
🚀 🚀

PR #1058, opened 2024-10-04
feat: global rules

hubertmis:
This pull request implements the idea sketched in https://github.com/likec4/likec4/pull/1038#issuecomment-2382135923. If merged, it would render #1038 obsolete.

It fixes two proposals from the discussion: #961 and #962.

The main benefits of the new features:

* Styles can be defined globally (global styles) and defined to each `view {}` by a reference
* Styles defined globally can be grouped (global style groups). Each `view {}` node can apply style groups
* Styles can be defined for a `views {}` block (local styles) and applied to each `view {}` in the block
* Styles (and groups) defined globally can be referenced in a `views {}` block and applied to each `view {}` node in the block
* Include and exclude predicates can be defined in global groups (global predicate groups) and applied to views

### DIFF
--- a/apps/docs/src/content/docs/dsl/views.mdx
+++ b/apps/docs/src/content/docs/dsl/views.mdx
@@ -469,6 +469,35 @@ Together with [extend views](#extend-views) you may define a "baseline" view (in
 
 :::
 
+#### Global predicates
+
+Predicates can be defined in global groups.
+Eache global group can be applied to multiple views.
+When a group is applied to a view, all predicates from the group are applied.
+
+```likec4
+global {
+  predicateGroup new_cloud_service {
+    include cloud.*
+      where kind is microservice
+    exclude *
+      where tag is #deprecated
+  }
+}
+
+views {
+  view of newServices {
+    include *
+    global predicate new_cloud_service
+  }
+
+  view of newBackendServices {
+    include *
+    global predicate new_cloud_service
+  }
+}
+```
+
 ### Auto-layout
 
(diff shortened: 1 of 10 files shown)


davydkov:
Yes, this is exactly what I meant! It seems that shared predicates might make 'view extends' obsolete (which is not perfect). I'll take a closer look tomorrow

hubertmis:
@davydkov , did you take a look at this proposal? Can we proceed with something like this?

davydkov:
Hey @hubertmis

Sorry, for late response.

I like this variant 🚀 🚀 🚀 It covers most use-cases!!

I suggest the following order (each step in separate PR):

- **Step 1:** Styles can be defined in `views {}` block (local styles) and applied to each `view {}` in the block (seems very easy, scope `model-parser`)

- **Step 2:** Styles can be defined globally (global styles) and applied inside `view {}` by a reference (scope `model-builder`)

- **Step 3:** Styles defined globally can be grouped (global style groups). Each `view {}` node can apply style groups

- **Step 4:** Styles (and groups) defined globally can be referenced in a `views {}` block and applied to each `view {}` node in the block
  And since order is important (to apply styles), in langium we need to change
  ```
  ModelViews:
    name='views' '{' (
        views+=LikeC4ViewRule |
        styles+=(ViewsLocalStyle | ViewsGlobalStyleRef)
      )*
  ```

hubertmis:
Ok, let's do this. I'll start opening PRs replacing this one

Issue #1065, opened 2024-10-07
User defined colors does not work


pavelpykhtin:
The issue seems to be related to Content Security Policy. Styles for user defined colors rely on 'unsafe-inline' rule of CSP. But in latest version of VSCode/extension another style block tagged with nonce was added. According to [CSP: style-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src) unsafe-inline is ignored if there is any style with nonce.

PR #1071, opened 2024-10-08
Fix for #1065

pavelpykhtin:
Added nonce parameter for marking dynamic styles inserted by components.

### DIFF
--- a/packages/vscode-preview/src/main.tsx
+++ b/packages/vscode-preview/src/main.tsx
@@ -21,11 +21,13 @@ import { theme } from './theme'
 
 const root = document.getElementById('root') as HTMLDivElement
 const scheme = document.body.classList.contains('dark') ? 'dark' : 'light'
+const nonce = root.getAttribute('nonce') || undefined
+
 // const reactRoot = ReactDOM.createRoot(root)
 ReactDOM.createRoot(root).render(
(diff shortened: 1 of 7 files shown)


packages/vscode-preview/src/main.tsx
```diff
@@ -21,11 +21,13 @@ import { theme } from './theme'
...
 const root = document.getElementById('root') as HTMLDivElement
 const scheme = document.body.classList.contains('dark') ? 'dark' : 'light'
+const nonce = root.getAttribute('nonce') || undefined
+
 // const reactRoot = ReactDOM.createRoot(root)
 ReactDOM.createRoot(root).render(
   <MantineProvider theme={theme} forceColorScheme={scheme}>
     <LikeC4Context>
-      <App />
+      <App nonce={nonce} />
```
davydkov: I just checked, `MantineProvider` has `getStyleNonce` property
And we can get it with `useMantineStyleNonce` hook

davydkov (changes requested):

packages/vscode-preview/src/main.tsx
pavelpykhtin: Good point! Thanks!

davydkov (approved):

PR #1099, opened 2024-10-11
Implemented alignment tools

pavelpykhtin:
Implemented alignment tools for manual layouting. Alignment applied to leaf nodes only.
alignment-tools

Also replaced @lume/kiwi solver with simple implementation. kiwi solver does not work properly for compound nodes on 2+ level with a single child. I guess that was the reason why dragging was disabled for nodes which are the only child. This workaround is acceptable for drag scenario but does not fit to align scenario: it is still required to align a node with others, even it is the only child.
Although implementation with solver is more flexible than bounding box implementation, this flexibility is not used.

### DIFF
--- a/packages/diagram/package.json
+++ b/packages/diagram/package.json
@@ -31,7 +31,6 @@
   },
   "dependencies": {
     "@likec4/core": "workspace:*",
-    "@lume/kiwi": "^0.4.4",
     "@mantine/vanilla-extract": "7.13.2",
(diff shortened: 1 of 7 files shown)


davydkov:
Hey @pavelpykhtin  🚀 🚀 🚀

I like the result!!

Just to explain. Initially, I picked `@lume/kiwi` with a thought to handle collisions, but that was quite tricky to get.

PR #1130, opened 2024-10-19
feat: global styles

hubertmis:
Define global styles that can be applied to any views in the document.

This is the second PR of a series aiming to replace #1058

### DIFF
--- a/packages/core/src/types/global.ts
+++ b/packages/core/src/types/global.ts
@@ -1,0 +1,10 @@
+import type { Tagged } from 'type-fest'
+import type { NonEmptyArray } from './_common'
+import type { ViewRuleStyle } from './view'
+
+export type GlobalStyleID = Tagged<string, 'GlobalStyleID'>
+
+export interface GlobalStyle {
+  readonly id: GlobalStyleID
+  readonly styles: NonEmptyArray<ViewRuleStyle>
+}
--- a/packages/core/src/types/model.ts
+++ b/packages/core/src/types/model.ts
@@ -1,4 +1,5 @@
 import type { ElementKindSpecification, Tag, TypedElement } from './element'
+import type { GlobalStyle, GlobalStyleID } from './global'
 import type { Relation, RelationID, RelationshipKindSpecification } from './relation'
 import type { ComputedView, DiagramView, LikeC4View, ViewID } from './view'
 
@@ -19,6 +20,9 @@ export interface ParsedLikeC4Model<
   }
   elements: Record<Fqns, TypedElement<Fqns, ElementKinds, Tags>>
   relations: Record<RelationID, Relation>
+  globals: {
+    styles: Record<GlobalStyleID, GlobalStyle>
+  }
   views: Record<Views, LikeC4View<Views, Tags>>
 }
 
--- a/packages/language-server/src/ast.ts
+++ b/packages/language-server/src/ast.ts
@@ -93,6 +93,10 @@ export interface ParsedAstRelation {
   metadata?: { [key: string]: string }
 }
 
+export interface ParsedAstGlobals {
+  styles: Record<c4.GlobalStyleID, c4.NonEmptyArray<c4.ViewRuleStyle>>
+}
+
 export interface ParsedAstElementView {
   __: 'element'
   id: c4.ViewID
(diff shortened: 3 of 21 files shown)


packages/core/src/types/global.ts
```diff
@@ -0,0 +1,9 @@
+import type { Tagged } from 'type-fest'
+import type { ViewRuleStyle } from './view'
+
+export type GlobalStyleID = Tagged<string, 'GlobalStyleID'>
+
+export interface GlobalStyle {
+  readonly id: GlobalStyleID,
+  readonly styles: ViewRuleStyle[],
+}
```
davydkov: Seems you always have array with single element
Should it be?

```suggestion
export interface GlobalStyle {
  readonly id: GlobalStyleID,
  readonly style: ViewRuleStyle,
}
```

packages/core/src/types/model.ts
```diff
@@ -19,6 +20,9 @@ export interface ParsedLikeC4Model<
   }
   elements: Record<Fqns, TypedElement<Fqns, ElementKinds, Tags>>
   relations: Record<RelationID, Relation>
+  globalRules: {
```
davydkov: I suggest rename to `globals`

packages/language-server/src/ast.ts
```diff
@@ -93,6 +93,10 @@ export interface ParsedAstRelation {
   metadata?: { [key: string]: string }
 }
 
+export interface ParsedAstGlobalRules {
+  styles: Record<c4.GlobalStyleID, c4.ViewRuleStyle[]>,
```
davydkov:
```suggestion
  styles: Record<c4.GlobalStyleID, c4.ViewRuleStyle>,
```

packages/language-server/src/validation/index.ts
```diff
@@ -31,6 +34,9 @@ export function registerValidationChecks(services: LikeC4Services) {
     IconProperty: iconPropertyRuleChecks(services),
     SpecificationRule: specificationRuleChecks(services),
     Model: modelRuleChecks(services),
+    ModelViews: modelViewsChecks(services),
```
davydkov: I removed this validation as now I see no reason to warn if there are multiple `views` blocks in a document, why not?

What do you think?

packages/language-server/src/model/model-builder.ts
```diff
@@ -211,6 +212,32 @@ function buildModel(services: LikeC4Services, docs: ParsedLikeC4LangiumDocument[
...
+  }
+  Object.assign(parsedGlobalRules.styles, ...docs.map(d => d.c4Global.styles))
+
+  const globalRules: {
+    styles: Record<c4.GlobalStyleID, c4.GlobalStyle>,
+  } = {
+    styles: {},
+  }
+
+  for (const key in parsedGlobalRules.styles) {
```
davydkov: I suggest replace with 
```suggestion
  for (const [id, styles] of entries(parsedGlobalRules.styles)) {
```
to remove redundant `if` clause in `for`

or even
```ts
globalRules.styles = pipe(
  entries(parsedGlobalRules.styles),
  map(([id,style]) => ({id, style})),
  indexBy(prop('id'))
)
```

packages/core/src/types/global.ts
hubertmis: For now your suggestion would be better, but I have already prepared these types for style groups coming in next PR. I would prefer to keep it instead of reworking twice if that's ok for you

davydkov: Ah, I see, indeed makes sense.
Maybe make it stricter with `NonEmptyArray`, but this is optional

packages/language-server/src/validation/index.ts
hubertmis: It looks like a rebase error. I didn't intend to restore it

packages/language-server/src/model/model-builder.ts
hubertmis: Thanks for this solution! I spent long time trying to figure out how to do this in TypeScript. Finally I gave up.

I'll get familiar with this library. Now I have no clue how it works. But the tests are passing :)

davydkov (approved):

PR #1143, opened 2024-10-21
feat: global style groups

hubertmis:
Allow defining style groups with a set of style rules that can be applied to multiple views.

This is the 3rd PR of a series aiming to replace #1058


davydkov:
🚀 🚀

Issue #1144, opened 2024-10-23
Not working syntax where relationship together with with


viktorgusev1984:
When used with relationship predicates where and together with with, not work style
example 
```
include
  cloud.* <-> amazon.*
    where kind is http
    with {
      color amber
    }
```

PR #1147, opened 2024-10-24
Fix on #1144

pavelpykhtin:
Fixes where condition being ignored by if it is followed by 'with' expression.


davydkov:
👍🏻 👍🏻 
Resolves #1144

PR #1149, opened 2024-10-25
Unique suffixes for items in defs sections of svg

pavelpykhtin:
@davydkov , linearGradient is defined in defs section of svg and thus is scoped to the whole document rather than that particular image. Added fix with unique suffix to defs content.

I'm not quite happy with the implementation. It feels wrong to do that many changes at once. Do you have some kind of import script for icons? Probably fix could be applied there?


davydkov:
Hey @pavelpykhtin 
Wow, great that you nailed the issue.

But these changes will be overwritten by [`npm run update-icons`](packages/icons/scripts/update-icons.mjs) - right? 

What if we utilise some plugin from svgo, like [here](https://github.com/svg/svgo/issues/674#issuecomment-1803604056)

pavelpykhtin:
Totally missed that script! Thanks for pointing out!

Added prefixIds plugin into config.

PR #1173, opened 2024-10-27
feat: global element and relation predicates

hubertmis:
This feature is intended for reusable include / exclude sequences

Architecture documentation can contain multiple views sharing the same set of repeated include and exclude sequences. Instead of repeating this set, the maintenance gets easier when the set is defined once and applied to all relevant views.

This feature allows for defining "global predicate groups" that can contain a list of include and exclude predicates. Each group can be explicitly applied to any view.

This is the 5th and final PR in the series replacing #1058

### DIFF
--- a/apps/docs/src/content/docs/dsl/views.mdx
+++ b/apps/docs/src/content/docs/dsl/views.mdx
@@ -469,6 +469,35 @@ Together with [extend views](#extend-views) you may define a "baseline" view (in
 
 :::
 
+#### Global predicates
+
+Element and relationship predicates can be defined in global groups.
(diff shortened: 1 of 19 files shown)


davydkov:
Hey @hubertmis 

If you wonder why it took almost one week to merge.
I realised that better to keep processing of global rules at same place as regular rules.
Thus, we ensure that `ComputeCtx` handles input with global rules.
Previously, builder prepared view (resolved global rules) before passing it to compute - though it was not enforced on type-level by `ComputeCtx`.

#1197 addresses this

PR #1190, opened 2024-10-30
fix: apply global styles in all building scenarios

### LINKED ISSUE #1180: Global styles don't work under 1.15
The following trivial example should produce a single red box:

```
specification {
  element thing
}

model {
  thing1 = thing
}

global {
    style make_red * {
      color red
    }
}

views {
  view thing1view of thing1 {
    include *

    global style make_red
  }
}
```

It does not. The box remains the default color (blue).

This was tested in the Playground but the same behavior is present in VSCode and the LikeC4 dev server (both v1.15.0)

hubertmis:
Fixes #1180

Global styles were applied only when computing views. Now they are applied both during computing views and building computed models. It should cover all view computing paths.


davydkov (approved):
👍🏻 👍🏻

Issue #1235, opened 2024-11-15
[Feature request] - Diagram search


davydkov:
From https://github.com/likec4/likec4/discussions/1223
<div type='discussions-op-text'>

<sup>Originally posted by **vedeja** November 11, 2024</sup>
When browsing my growing collection of diagrams I often have a particular keyword in my head. In Structurizr we can hit the space bar and start typing. Could we have something similar @davydkov?</div>

Consider:
- [ ] In browser full-text-search, something like [flexsearch](https://github.com/nextapps-de/flexsearch) 
- [x] [Spotlight](https://mantine.dev/x/spotlight/#scrollable-actions-list) from Mantine or another?
- [ ] VSCode extension action, like "Go to view"

Cyclonit:
To extend on this, it would be great if searching for something would highlight all matches or preferably fade out elements that do not match. For example, we add communication protocols and encryption standards to edges' metadata. I'd love if searching for "protocol:HTTP" would make those edges pop whereas other edges are faded out.

PR #1244, opened 2024-11-20
Diargram search (#1235)

pavelpykhtin:
Implemented search for views and nodes in current view with @mantine/spotlight component.

Please, check that I've covered all export scenarios (static site, webcomponent etc.). I suspect that I could miss something.

### DIFF
--- a/apps/playground/package.json
+++ b/apps/playground/package.json
@@ -27,6 +27,7 @@
     "@likec4/log": "workspace:*",
     "@mantine/core": "7.13.4",
     "@mantine/hooks": "7.13.4",
+    "@mantine/spotlight": "7.13.4",
     "@mantine/vanilla-extract": "7.13.4",
     "@react-hookz/web": "^24.0.4",
     "@tabler/icons-react": "3.17.0",
--- a/apps/playground/src/main.tsx
+++ b/apps/playground/src/main.tsx
@@ -2,6 +2,7 @@
 import './font-ibm-plex-sans.css'
 import './font-fira-code.css'
 import '@mantine/core/styles.css'
+import '@mantine/spotlight/styles.css'
 import './main.css'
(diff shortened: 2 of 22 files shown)


Cyclonit:
Hi @pavelpykhtin,

I am not a contributor/maintainer myself, but I gave it a try and like it a lot. I didn't find any problems in the VS Code extension at least.

What do you think about adding a search button to the TopLeftPanel, that opens the Spotlight? Something like this could do. I am very new to React and didn't find a way of including the shortcut in the tooltip (besides hardcoding it in two places). That would be the cherry on top.

```html
<Tooltip label="Search">
  <ActionIcon
    onClick={() => openSpotlight()}>
    <IconSearch />
  </ActionIcon>
</Tooltip>
```

Cyclonit:
Another improvement would be to include the part of the element that matches the search query in the `Action`. When the query matches anything other than the title/label (e.g. a tag), it is not clear why it shows up in the results. Maybe it'd be possible to include a short dynamic hint after/below the action title, that shows why the action is being shown.

Maybe like this
grafik

Cyclonit:
I prepared a little sample based on your work here: https://github.com/likec4/likec4/compare/feat/diagram-search...Cyclonit:likec4:feat/diagram-search. What do you think about it?

davydkov:
> I prepared a little sample based on your work here: [feat/diagram-search...Cyclonit:likec4:feat/diagram-search](https://github.com/likec4/likec4/compare/feat/diagram-search...Cyclonit:likec4:feat/diagram-search). What do you think about it?

Hey @Cyclonit 
I like the idea!
Can you check this https://mantine.dev/core/highlight/?
Seems a perfect fit

davydkov:
@pavelpykhtin Nice one! Thank you!

PR #1250, opened 2024-11-24
Deployments model

davydkov:
Work in progress

- [x] Specification for deployment kinds
- [x] Deployment nodes
- [x] `instanceOf`
- [x] Deployment view
- [x] Deployment model relationships
- [ ] Deployment view predicates:
  - [x] Nodes and instances 
  - [ ] Relationships

Playground: https://432f4a58.likec4-playground.pages.dev/w/deployment/index/

Locally - checkout branch and launch "Run Extension" in VSCode

Discussion - https://github.com/likec4/likec4/discussions/1269


Cyclonit:
This looks dope already, very nice work. I'll allocate some time early next week to test it using some examples from work, such that I can give proper feedback.

One question though, is it possible for the system to support intersecting groups? For example, container cluster might span multiple availability zones in AWS, but does not belong to any AZ itself. It would be great to visualize this by groups like this:

grafik

davydkov:
> 
> One question though, is it possible for the system to support intersecting groups?
> 
> 

My initial thought is that it should be possible, though it might be tricky to implement with Graphviz.
And another open question - should it be represented in model? Or view only? And how?

Cyclonit:
I can only think of use cases for having it be part of the deployment view. But I cannot say for certain that there aren't some for having it in the model too.

Kiiv:
Hi !

Wow, thanks for the work. It's already a really good work in progress and it probably cover a lot of use case.

However, one thing I'm not really comfortable with is to use the exclusion capabilities of the views to represent something inherent to the model.
I mean, the fact that two zones can't communicate is part of the deployment itself. Just hiding relationships with exclusion does not make this information a part of the design.
Maybe I'm wrong but for me a view is just a representation of the model, it should not be used to define it.
For example, if I build an other view containing those deploymentNodes, I will need to do the exclusion again. If different people work on the model, it could not be clear if it's an exclusion just for this view or if it need to be reported everywhere.

Do you think it could be possible to add an attribute to an instanceOf to make them part of a cluster ?
Instances of a cluster could communicate with instances of the same cluster and instances with no cluster defined :

```
  environment prod 'Production' {
    eu = zone 'EU' {
      zone zone1 {
        instanceOf saas.ui {
}  environment prod 'Production' {
    eu = zone 'EU' {
      zone zone1 {
        instanceOf saas.ui {
          cluster cluster1
        }
        instanceOf saas.backend {
          cluster cluster1
        }
      }
      zone zone2 {
        instanceOf saas.ui {
          cluster cluster2
        }
        instanceOf saas.backend {
          cluster cluster2
        }
      }
    }

    us = zone 'US' {
      zone zone1 {
        instanceOf saas.ui {
          cluster cluster3
        }
        instanceOf saas.backend {
          cluster cluster3
        }
      }
      zone zone2 {
        instanceOf saas.ui {
          cluster cluster4
        }
        instanceOf saas.backend {
          cluster cluster4
        }
      }
    }
```

Maybe "cluster" attribute should be a list.

Cyclonit:
### Disallowing Relationships

If I understand you @Kiiv correctly, the problem statement is as such:

> Deployments may alter relationships between instances of elements that are defined in the model. For example, in a distributed system, cross-region traffic from a frontend in region A might not be allowed to traverse to backends in region B.

I do not like the use of the term "cluster" here. It is specific to a specialization of the problem and does not properly map to most cases. Also, semantically speaking, there is nothing inherent to a cluster that should prevent two distinct clusters from talking to each other. But I do agree, that "deployment overrides" like this should be part of the model and not the view.

For the case of removing relationships, what about using the existing `exclude` keyword from views and making it part of the `deployment` instead? I believe the following makes perfect sense:

```
deployment {
    // disallow all relationships between regions and zones
    exclude
      eu._ <-> us._,
      eu.zone1._ <-> eu.zone2._,
      us.zone1._ <-> us.zone2._,
}
```

### Intercepts

I worded the problem statement more general by saying "alter" instead of "disallow", because a deployment may introduce further alterations over the functional modal like introducing load balancers. As far as I can tell, there is not such construct in this PR yet. What do you think about this?

```
specification {
    element loadbalancer
}

deployment {

  node customerNode {
    // References to model.c4
    instanceOf customer
  }

  environment prod 'Production' {
    eu = zone 'EU' {
      lb = loadbalancer

      zone zone1 {
        instanceOf saas.ui
        instanceOf saas.backend
      }
      zone zone2 {
        instanceOf saas.ui
        instanceOf saas.backend
      }

      intercept saas.customer -> lb -> saas.ui
    }

    ....
}
```

Things I like about an approach like this:

1. It should be possible to define elements that are exclusive to the deployment model in the deployment model. The functional model should not care about load balancers and the like. Thus having them in the deployment model is ideal.
2. The simple syntax of `intercept <from> -> <via> -> <to>` is easily understood.

Things I do not like:

1. While an `intercept` encapsulates the idea of the load balancer being put in the middle of an existing relationship, it makes adding labels to the newly created edges difficult.

Cyclonit:
> > One question though, is it possible for the system to support intersecting groups?
> 
> My initial thought is that it should be possible, though it might be tricky to implement with Graphviz. And another open question - should it be represented in model? Or view only? And how?

I see it is becoming very complicated fast :D

I would argue that these intersecting groups are neither elements in the model nor nodes in the deployment. But they might make sense in both. I see two options:

1. Introduce a new kind of element called `group`. A group cannot have children defined inside of it, but it may include elements/nodes defined elsewhere:

```
deployment {
  frontend_eu = group 'Frontend Cluster EU' {
    include eu.*.saas.ui
  }
  frontend_us = group 'Frontend Cluster US' {
    include us.*.saas.ui
  }
}
```

2. Instead of defining a new group construct, extend the existing element construct with an ability to include elements defined elsewhere.

Kiiv:
> ### Disallowing Relationships
> 
> I do not like the use of the term "cluster" here. It is specific to a specialization of the problem and does not properly map to most cases. Also, semantically speaking, there is nothing inherent to a cluster that should prevent two distinct clusters from talking to each other. But I do agree, that "deployment overrides" like this should be part of the model and not the view.

Totally agree about the term "cluster". Thinking about it this morning and I was planning to edit my previous comment about it :)

> For the case of removing relationships, what about using the existing `exclude` keyword from views and making it part of the `deployment` instead? I believe the following makes perfect sense:
> 
> ```
> deployment {
>     // disallow all relationships between regions and zones
>     exclude
>       eu._ <-> us._,
>       eu.zone1._ <-> eu.zone2._,
>       us.zone1._ <-> us.zone2._,
> }
> ```

One thing I'm not totally comfortable with is to rely on the deploymentNode hierarchy to limit relationships. It could be an easy way but I have a use case like this (simplified) :

LikeC4_deployment_example

Some instances are dedicated to one kind of user and some others to an other kind. I don't have "zones" to represent the dependencies between instances.
As you can see, I'm able to represent it with exclusions too but it's not straightforward. If more than two instances are implicated it could become a long list of exclusions.
That's why I though in the first place that a dedicated attribute could be relevant to explicitly "group" some instances.

But I could live with the exclusion if you think it's the best and most flexible solution (I can understand that adding new attributes for each use case could be cumbersome) :)


> ### Intercepts
> I worded the problem statement more general by saying "alter" instead of "disallow", because a deployment may introduce further alterations over the functional modal like introducing load balancers. As far as I can tell, there is not such construct in this PR yet. What do you think about this?

I pretty like it ! Seems really easy to use and it is excplicit :)
Maybe for the last point you don't like, it's just possible to use custom relationship ?
```
       intercept saas.customer -[toLB]-> lb -[fromLB]-> saas.ui
```

PR #1251, opened 2024-11-24
Highlight in Diagram Search

Cyclonit:
Follow up to #1244.

1. Add search button to the TopLeftPanel.
2. Add highlighting of matched keywords to diagram search.

GIF 24-11-2024 07-29-19

### DIFF
--- a/packages/diagram/src/LikeC4Search.tsx
+++ b/packages/diagram/src/LikeC4Search.tsx
@@ -1,83 +1,146 @@
-import { Spotlight, type SpotlightActionGroupData } from '@mantine/spotlight'
-import { IconRectangularPrism, IconSearch, IconSitemap } from '@tabler/icons-react'
-import { memo, useMemo } from 'react'
+import type { DiagramView } from '@likec4/core'
+import { Spotlight, SpotlightActionsGroup, type SpotlightActionData } from '@mantine/spotlight'
+import { IconRectangularPrism, IconSitemap, IconSearch } from '@tabler/icons-react'
+import { memo, useMemo, useState } from 'react'
 import { filter, map, pipe } from 'remeda'
 import { useDiagramState, useDiagramStoreApi, useMantinePortalProps } from './hooks'
 import { useLikeC4Model } from './likec4model'
+import { Center, Group, Highlight, Text } from '@mantine/core'
+
+interface QueryMatch {
+  data: SpotlightActionData,
+  keyword: string,
+  match: string
+}
+
+function filterAction(actionData: SpotlightActionData, query: string) : QueryMatch | undefined {
+  const queryLower = query.toLowerCase();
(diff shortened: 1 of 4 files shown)


packages/diagram/src/LikeC4Search.tsx
```diff
@@ -1,83 +1,142 @@
...
+      leftSection: <IconSitemap />
+    }));
   }
 
-  const actions: SpotlightActionGroupData[] = useMemo(() => [
-    getNodeActions(),
-    getViewActions()
-  ], [model, store, view])
+  const nodeActions = buildFilteredActions(useMemo(getNodeActionsData, [model, store, view]), query);
+  const viewActions = buildFilteredActions(useMemo(getViewActionsData, [model, store, view]), query);
```
davydkov: What about wrapping in single `useMemo`, like:

```suggestion
  const { nodeActions, viewActions } = useMemo(() => ({
      nodeActions: buildFilteredActions(getNodeActionsData, query),
      viewActions: buildFilteredActions(getViewActionsData, query)
    }),
    [model, store, view, query]
  )
```

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
```diff
@@ -174,6 +176,13 @@ export const TopLeftPanel = () => {
         gap={'xs'}>
         {showNavigationButtons && <BackwardForwardButtons />}
         <ActionIconGroup className={css.actionIconGroup} orientation="vertical">
+          <Tooltip label="Search">
```
davydkov: I think we need to pass `enableSearch` from props to store, like:

https://github.com/likec4/likec4/blob/3c350b44f3602eaa3aaede2d57ff0e725b3d5c22/packages/diagram/src/LikeC4Diagram.tsx#L139

And show this icon if flag is set.

davydkov (changes requested):
Hey @Cyclonit, great job!
Congratulations on your first contribution.

I would ask you to add `enableSearch` to store, add to `DiagramInitialState` and update `LikeC4Diagram`

https://github.com/likec4/likec4/blob/3c350b44f3602eaa3aaede2d57ff0e725b3d5c22/packages/diagram/src/LikeC4Diagram.tsx#L115-L118

I think we can remove prop from `LikeC4DiagramInnerMemo`, as we can read from store

packages/diagram/src/LikeC4Search.tsx
Cyclonit: Changed as requested.

packages/diagram/src/ui/top-left/TopLeftPanel.tsx
Cyclonit: I think I did what you wanted. I am not entirely certain whether I looked in the right places though.

davydkov (approved):

Issue #1105, opened 2024-10-14
Render description and link in relationship details popup


gparmeggiani:
Since [v1.11.0](https://github.com/likec4/likec4/releases/tag/v1.11.0) we got the fantastic popup view with the relationship details.

It would be great if this was showing also the relationship description and link. And it would complete https://github.com/likec4/likec4/issues/612

PR #1307, opened 2024-11-29
Description and links are added to the context menu of a relationship

pavelpykhtin:
Finishes #1105


davydkov:
Looks good, I like it!

Issue #1259, opened 2024-11-24
Predicate to include all descendants `.**`


davydkov:
I've been thinking about possible options for such a predicate. Glob patterns use `**` for this purpose. Thus `saas.**` would include all descendants of `saas`. Is there any technical reason for why this should not exist? If not, I'd like to give implementing it a shot.

_Originally posted by @Cyclonit in https://github.com/likec4/likec4/discussions/1114#discussioncomment-11361687_

Cyclonit:
Hi @davydkov,

I am (slowly) working my way through the language definition and how the parsing works. If I understand everything correctly, `*` on "root" level and `element.*` are two fundamentally different constructs. One is a so called "wildcard" and the other sets the flag `ElementRef.isDescendants`. I will focus on the later for now, but I have a question regarding the former.

According to [/apps/docs/src/content/docs/dsl/views.mdx:160+](https://github.com/likec4/likec4/blob/main/apps/docs/src/content/docs/dsl/views.mdx?plain=1#L169) `*` behaves differently when used inside a `view` compared to a `view of`. To me, it looks like the behavior in `view of` is exactly what `**` is supposed to achieve in the future. Would it make sense to introduce a top level `**`, that behaves like `*` for scoped views, but works for both scoped and unscoped views? At that point, `*` and `**` would both work in scoped views. A second (albeit breaking) change would be to change `*` in scoped views to behave exactly like it does in unscoped views.

Example:
```
views {
  // old: include all top level elements
  view {
    include *
  }
  // new: include all top level elements and their descendants -> all elements
  view {
    include **
  }
  // old: include cloud.ui and all of its descendants
  view of cloud.ui {
    include *
  }
  // new: include cloud.ui and all of its descendants
  view of cloud.ui {
    include **
  }
}
```

On a side note, I would rename `isDescendants` to `isChildren`, such that `isDescendants` is available for the new selector. Would there be any reason not to do that?

davydkov:
> looks like the behavior in `view of` is exactly what ** is supposed to achieve in the future. 

`view of smth` has several semantics.
First, inside scoped views predicate `include *` becomes like:
```zig
include smth
include smth.* // includes only children
include -> smth.* -> // includes incoming to/outgoing from "outside" of smth
```
Second - a scoped view inherits the “scope” of the associated element.
Third - the scoped view becomes the default for the element.

I’m skeptical about introducing a predicate `include **` without further exploration of its use cases.
[...]

PR #1310, opened 2024-11-29
#1259 Predicate to include all descendants `.**`

Cyclonit:
Add a new predicate `.**` to include all of an element's descendants into a view. 

Example:
```
views {
  view index {
    include.saas.**
  }
}
```

@davydkov, where do you want me to update the documentation? After the current documentation for `.*` might be a good place.

### DIFF
--- a/packages/core/src/builder/Builder.view.ts
+++ b/packages/core/src/builder/Builder.view.ts
@@ -311,12 +311,12 @@ function $expr<Types extends AnyTypes>(expr: ViewPredicate.Expression<Types> | C
   if (expr.endsWith('.*')) {
     return {
       element: expr.replace('.*', '') as Fqn,
-      isDescedants: true
+      isChildren: true
     } as TypedC4Expression<Types>
   }
   return {
     element: expr as any as Fqn,
-    isDescedants: false
+    isChildren: false
   } as TypedC4Expression<Types>
 }
 
--- a/packages/core/src/builder/Builder.view.ts
+++ b/packages/core/src/builder/Builder.view.ts
@@ -314,9 +314,15 @@ function $expr<Types extends AnyTypes>(expr: ViewPredicate.Expression<Types> | C
       isChildren: true
     } as TypedC4Expression<Types>
   }
+  if (expr.endsWith('.**')) {
+    return {
+      element: expr.replace('.**', '') as Fqn,
+      isDescendants: true
+    } as TypedC4Expression<Types>
+  }
   return {
     element: expr as any as Fqn,
-    isChildren: false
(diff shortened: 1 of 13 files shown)


packages/core/src/builder/Builder.view.ts
```diff
@@ -311,12 +311,18 @@ function $expr<Types extends AnyTypes>(expr: ViewPredicate.Expression<Types> | C
...
+  if (expr.endsWith('.**')) {
+    return {
+      element: expr.replace('.**', '') as Fqn,
+      isDescendants: true
     } as TypedC4Expression<Types>
   }
   return {
     element: expr as any as Fqn,
-    isDescedants: false
+    isChildren: false // TODO: why do we set this to false? Shouldn't undefined work too?
```
davydkov, line 325: Yes, you are right, no reason to set it

packages/core/src/types/expression.ts
```diff
@@ -13,7 +13,9 @@ interface BaseExpr {
   elementKind?: never
   elementTag?: never
   isEqual?: never
-  isDescedants?: never
+  isChildren?: never
+  isDescendants?: never
+  isLeafs?: never
```
davydkov, line 18: I don't see it is used

packages/language-server/src/model-graph/compute-view/predicates.ts
```diff
@@ -266,15 +273,12 @@ function resolveElements(this: ComputeCtx, expr: Expr.ElementExpression): Elemen
     return nonexhaustive(expr)
   }
 
-  if (this.root === expr.element && expr.isDescedants !== true) {
+  // TODO: Why do we include the root's children in this?
+  if (this.root === expr.element && expr.isChildren !== true && expr.isDescendants !== true) {
```
davydkov, line 277: Oh, it’s a corner case, a bit of “sugar” for relation predicates, like:

```zig
view of element {
  include
     *,
     // Allows to write only this
     other -> element,

     // And works same as
     other -> element,
     other -> element.*,

     // But this has stricter semantics
     other -> element.*,
}
```

It was before `expand` predicate (that does the same), and I remember it was always confusing to write twice.

Maybe it makes sense to leave an explanation comment here; I believe it’s widely used.

davydkov (commented):
Such a valuable contribution and just 

image

Well done! 👏🏻 👏🏻 👏🏻 

> After the current documentation for .* might be a good place.

Yes, perfect.
Let's merge this and add docs in separate PR

Cyclonit:
Uhm @davydkov, did you merge this on accident? I didn't mark this as a draft, but my TODO comments should not have made it in :D

davydkov:
Oh sorry, you right, I was too excited and impatient 😄

PR #1340, opened 2024-12-09
refactor: extract component BottomButtons from all variants of ElementNode

Cyclonit:
**This is work in progress!**

Currently, the codebase contains a lot of duplicated and more importantly almost-but-not-really-duplicated code. This PR is intended as an entry into getting rid of some of this baggage. It extracts the "bottom buttons" in the various ElementNodes into a dedicated component.

I'd appreciate input on the remaining to dos. I am fairly new to react and don't quite understand where the dependencies are coming from/breaking yet.

**To do:**
- Fix button pop out alignment on hover (all buttons transition to the bottom left, instead of outwards in opposite directions)
- Fix button pop out animation not running on hover in relationships-of overlay and edge-details overlay.


Cyclonit:
Hi @davydkov, @pavelpykhtin,

I'd like some feedback on what I am doing. The overall goal is to simplify the codebase by getting rid of duplicate code. One part I found was the type definitions for xyflow nodes/edges. There are two almost identical definitions for these types:

- `packages/diagram/src/overlays/edge-details/_types.ts`
- `packages/diagram/src/overlays/relationships-of/_types.ts`

The variant in `relationships-of` mostly has additional properties for animations. Thus I started merging them and created a new version in [`packages/diagram/src/xyflow/_types.ts`](https://github.com/likec4/likec4/blob/2ecf5429713394245e0b8366953e9795525e7d29/packages/diagram/src/xyflow/_types.ts). I think all of the properties used by `relationships-of` can safely be put into the common base class. Currently, the animations are limited to `relationships-of`, but adding them to the other views would be nice anyways.

However, there is one exception:
```ts
type RelationshipsOfNodeProps = {
  column: 'incomers' | 'subjects' | 'outgoers'
}
```

The column property is needed for placement in the `RelationshipsOverlay`. It has no other use and thus should not reside in the common types declaration. However, I find my current solution in [`packages/diagram/src/overlays/relationships-of/_types.ts`](https://github.com/likec4/likec4/blob/2ecf5429713394245e0b8366953e9795525e7d29/packages/diagram/src/overlays/relationships-of/_types.ts) quite bad. Is there a better approach to doing something like that?

PR #1347, opened 2024-12-12
refactor: overlay deduplication

Cyclonit:
The goal of this PR is to reduce the amount of duplicate code in the codebase and to get to a more maintainable state in general. I am focusing on the overlays for now, because they are very similar.

I consider this to be the first in a chain of PRs. It addresses the types for xyflow backing the overlays. Prior to this PR, each overlay defines its own set of types. These types share some common identical properties (albeit defines separately), but also differ in some key points.

### Example
The properties `fqn` and `element` are identical between the two definitions. `existsInCurrentView` and `hovered` exist in only one of them, whereas `ports` has conflicting definitions.

**relationships-of/_types.ts**
```ts
type NodeProps {
  fqn: Fqn,
  element: Pick<ComputedNode, 'color' | 'title' | 'description' | 'shape' | 'kind'>
  existsInCurrentView: boolean
  ports: {
    left: Port[]
    right: Port[]
  }
}
```

**edge-details/_types.ts**
```ts
type NodeProps {
  fqn: Fqn,
  element: Pick<ComputedNode, 'color' | 'title' | 'description' | 'shape' | 'kind'>
  hovered: boolean
  ports: {
    in: string[]
    out: string[]
  }
}
```

### Solution

I extracted common properties into shared base types in the new [`packages/diagram/src/overlays/shared/xyflow/_types.ts`](https://github.com/likec4/likec4/blob/6f4828452c71d116bb7ac2ae7f773dd101b15006/packages/diagram/src/overlays/shared/xyflow/_types.ts). I included the properties used for animations (`hovered`, `leaving`, `entering` and `dimmed`) in the common base class, such that we can standardize animations in a follow up PR. This means we have one source of truth for these properties and are less prone to diverging implementations in the future.

Here is a list of direct consequences that have made it into this PR:
1. The overlay specific `_types.ts` no longer define their own `Node` and `Edge` types from the ground up. Instead they extend the types from `SharedTypes` with only those properties that are unique to them.
2. The handling of `ports` has been standardized. Before some places simply used a list of target/source ids as strings, while others used objects including the id and the type (`in`/`out`). Now all places use two lists of ids.
3. By getting rid of diverging code, code duplication in other places is becoming more and more obvious. For example, the handling of ports in all `ElementNode.tsx` and `CompoundNode.tsx` is clearly visible.

### Minor Changes
I also included a couple of minor changes, that are aimed and simply cleaning up some stuff.

1. Rename `includedInCurrentView` to `existsInCurrentView` for edges in `relatioonships-of` to align the wording with the same property in nodes.
2. Add comments for all shared types' properties.
3. Standardize handling of relations between `edge-details` and `relationships-of` overlays.

### DIFF
--- a/packages/diagram/src/overlays/edge-details/EdgeDetailsXYFlow.tsx
+++ b/packages/diagram/src/overlays/edge-details/EdgeDetailsXYFlow.tsx
@@ -20,6 +20,7 @@ import { CompoundNode } from './xyflow/CompoundNode'
 import { ElementNode } from './xyflow/ElementNode'
 import { RelationshipEdge } from './xyflow/RelationshipEdge'
 import type { SharedTypes } from '../shared/xyflow/_types'
+import { only } from 'remeda'
 
 const nodeTypes = {
   element: ElementNode,
@@ -195,8 +196,10 @@ export const EdgeDetailsXYFlow = memo<{ edgeId: EdgeId }>(function EdgeDetailsXY
       // }}
       onEdgeClick={(e, edge) => {
         e.stopPropagation()
+        // edge-details only ever deals with edges that represent only one relation
+        let relationId = only(edge.data.relations)!.id;
         diagramStore.getState().onOpenSource?.({
-          relation: edge.data.relationId
+          relation: relationId
         })
       }}
       // onEdgeClick={(e, edge) => {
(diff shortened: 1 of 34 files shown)


packages/diagram/src/overlays/edge-details/EdgeDetailsXYFlow.tsx
```diff
@@ -194,8 +196,10 @@ export const EdgeDetailsXYFlow = memo<{ edgeId: EdgeId }>(function EdgeDetailsXY
       // }}
       onEdgeClick={(e, edge) => {
         e.stopPropagation()
+        // edge-details only ever deals with edges that represent only one relation
+        let relationId = only(edge.data.relations)!.id;
```
davydkov:
```suggestion
        let relationId = only(edge.data.relations)?.id;
        if (relationId) {
```

davydkov (approved):
I like it, thank you! I should have done it sooner.
`AbstractRelation` indeed might be useful in `deriveEdgePropsFromRelationships`, `RelationshipModel` and `DeploymentRelationModel`

packages/diagram/src/overlays/edge-details/EdgeDetailsXYFlow.tsx
Cyclonit: implemented

Issue #1369, opened 2024-12-19
Implement where predicates on participants of relation in element views


pavelpykhtin:
It's a subtask for #1363 

A relation predicate must provide the ability to filter relations by the properties of its participants.

### Limitations
- This task covers only Element Views. Similar functionality for Deployment Views will be implemented in a separate task.

### Suggested syntax
```
view test {
    title 'Test view'

    include
      customer,
      -> * where target.kind == component
	}
```

PR #1371, opened 2024-12-19
Implemented predicates on participants of the relation for element views

pavelpykhtin:
Implementation for #1369


pavelpykhtin:
Hi, @davydkov !

I've fixed conflicts and added examples. Should be ready to merge.

PR #1379, opened 2024-12-24
Where conditions on relations in deployment views

### LINKED ISSUE #1370: Implement where predicates on relations in deployment views
It's a subtask for #1363 

- Deployment views must support filtering of relations with 'where' clause the same way as it is implemented in Element Views (filtering by tags and kind of relation)
- Deployment views must support customization of relation's properties with 'with' clause the same way as it is implemented in Element Views

pavelpykhtin:
Implementation of where conditions on relations in deployment view

Follow up to #1381 
Resolves #1370


davydkov:
Hey @pavelpykhtin!

Sorry I didn’t reach out sooner. I was away on holidays.

This PR is super important, but I made some changes that caused conflicts. But on the bright side, it seems like your changes are pretty easy to move around.

I would only suggest the same approach as filtering in the element view:
https://github.com/likec4/likec4/blob/59cdc655d37885b92dfe64ca67a8700224459f91/packages/core/src/compute-view/element-view/_types.ts#L31-L36

pavelpykhtin:
Hi, @davydkov ,

I'm working on rebase of my changes on top of the main branch. Right now I'm fixing tests and need some guidance on elements/explicits/implicits fields of the stage/memory. It used to follow logic: explicits are always visible unless they are excluded. But in the current main StageInclude.commit calculates explicits as an intersection of elements and union of staged and existing explicits (https://github.com/likec4/likec4/blob/b6a69ce283ead0c20fbdc1614c12256ae54fcdb4/packages/core/src/compute-view/memory/AbstractStageInclude.ts#L181). Thus, we could lose explicits added before. I guess, in most cases it's not a problem because explicits are automatically added to elements collection and this problem is only observed in tests. But was this the intended implementation, or just a hidden issue?

davydkov:
Hey @pavelpykhtin 

> But in the current main StageInclude.commit calculates explicits as an intersection of elements and union of staged and existing explicits. 

This is intentional.
`state.elements` is an accumulator that holds “everything”, including both explicit and implicit elements.

It’s important to note that:
- A JS Set maintains insertion order, so `state.elements` is a set of elements in the order they were included.
- The `intersection` function preserves the order from the first set.

In other words:
```ts
state.explicits = intersection(
  // preserve the order from this set
  state.elements,
  // take these elements
  union(
    state.explicits,
    this.explicits,
  ),
)
```

This ensures that elements, that were previously implicit but become explicit, still maintain their “order” as they were included before. This is crucial for the `topologicalSort` algorithm.

This ensures that element, previously implicit but becoming explicit, keeps its "order" as was included before. Order is important for `topologicalSort`

> Thus, we could lose explicits added before.

No, we cannot.
The first operation is a union of staged elements (`stage._ordered`) and memory elements, which includes explicit elements added before.

No, we can not.
First operation is a union of memory elements and staged elements (`stage._ordered`), and set `memory.elements` includes explicits added before.

The subsequent intersections just "pick" elements in the original order.
As a bonus, we ensure that `state.explicits` and `state.final` are subsets of `state.elements`.

>  this problem is only observed in tests.

And this is confusing. Could you share the test that demonstrates this problem?

pavelpykhtin:
@davydkov , thank you for detailed explanation!

The test in which I found this issue builds memory state with Memory.update method. It used to populate explicits collection and leave elements collection empty. It's not the best design of a test (too much knowledge about the internals of the class being tested). But it is still required to know internal structure of Memory class to verify results and using stage methods does not provide enough control over initial state of the memory (issue in any of add*, exclude* or commit methods will brake the test although the target method of the test may still work as expected).
Anyway, it's a problem of init stage of the test and not the problem of the existing code. I've fixed the test.

davydkov:
@pavelpykhtin
Ahh, I see, `memory.update` accepts partial patch, which may result in an invalid state.

However, it is the stage’s responsibility to ensure consistency and mutate memory.

Let me explain.

From the end user’s perspective, a predicate is a single, atomic operation.
However, under the hood, it was not.
The predicate directly mutated memory, which sometimes led to unexpected results.

Now, the predicate pushes mutations to the stage. The stage “squashes” all mutations and commits them to memory as a single operation.

Though, order of mutations is still important for `include`.
For instance, [`stage.connectWithExisting`](https://github.com/likec4/likec4/blob/main/packages/core/src/compute-view/element-view/memory/stage-include.ts#L15). First it creates connections from existing elements to the new one, and then from the new element to existing.

It maintains natural order of connections, and `topologicalSort` works better.
Previously, I tried different heuristics and sorting, but, in the end, the natural order gives the best results. (order is important for graphviz)


It maintains the natural order of connections, and `topologicalSort` performs better. Previously, I experimented with various heuristics and sorting methods, but keeping the natural order yields the best. (Order is crucial for Graphviz.)

pavelpykhtin:
Yes, I got the main idea of the stage. It's a really good improvement! 

I've completed rebase. During the testing I found an issue #1414. It's better to fix after merge to avoid another long round of 'rebase-fix'.

PR #1384, opened 2024-12-28
Fixed relation being yielded twice (#1364)

### LINKED ISSUE #1364: Relationship count seems erroneous in "Relationship view"
From this example : https://template.likec4.dev/view/cloud

In "current view" mode, all relationships have a label "2 relationships".
In "Global" mode, we can see that there is only one relationship between nodes.

Is it working as intended ?

pavelpykhtin:
Closes #1364


davydkov:
🦅

PR #1428, opened 2025-01-09
Restored code lost during rebase

pavelpykhtin:


davydkov:
Ouch!!
At least now we have specs 😄

PR #1430, opened 2025-01-10
Feat/1427 deployment rel participant filter

### LINKED ISSUE #1427: Implement where predicates on participants of relation in deployment views

pavelpykhtin:
Follow up to #1379 and #1428

Closes #1427

### DIFF
--- a/apps/docs/src/content/docs/dsl/Deployment/views.mdx
+++ b/apps/docs/src/content/docs/dsl/Deployment/views.mdx
@@ -74,12 +74,10 @@ When condition on relation participant is checked the following rules are applie
 - For a deployment instances tags are combined from tags defined in model and tags defined in deployment model
 - For a child of deployment instance the tags defined on child in model are used
 - For a deployment node the tags defined in deployment model are used
+- For a deployment instances the kind of the model element is used
+- For a child of deployment instance the kind of this child is used
+- For a deployment node the kind of this deployment node is used
 - Tags are not inherited from parent nodes/elements
-
-
-<Aside type='caution' title='In development'>
-Filtering by deployment node kind is not implemented
-</Aside>
 
 ```likec4
 model {
--- a/apps/docs/src/content/docs/dsl/Deployment/views.mdx
+++ b/apps/docs/src/content/docs/dsl/Deployment/views.mdx
@@ -54,18 +54,80 @@ But they refer to deployment nodes and instances.
 
   ```likec4
   deployment view prod {
-    include *                            // works
-    include * where tag is #next         // does not work
-    include * with { color: red }        // does not work
-      
-    include * -> *                       // works
-    include * -> * where tag is #next    // works
-    include * -> * with { color: red }   // does not work
+    include *                                   // works
+    include * where tag is #next                // does not work
+    include * with { color: red }               // does not work
 
-    global style applications            // does not work    
+    include * -> *                              // works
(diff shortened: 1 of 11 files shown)


pavelpykhtin:
Hi, @davydkov ,

Resolving properties of the deployment nodes is a bit trickier than the same task for the model elements.
Deployment instance can have its own tags and kind.

So far, I came up with the following rules:
1. For element refs tags must be matched with the element's tags
2. For deployment instance refs tags must be matched with the union of the instance tags and its model element's tags
3. For deployment node refs tags must be matched with the deployment node's tags
4. For element refs and deployment instance refs kind must be matched with the element's kind
5. For deployment node refs kind must be matched with the deployment node's kind

Do you agree with these rules?

I'm not sure in meaning of the NestedElementOfDeployedInstanceModel. It has instance and element properties. Does it describe an element defined inside the element from which the instance is created (defined)?

davydkov:
Hey @pavelpykhtin 

> So far, I came up with the following rules:
> 
> 1. For element refs tags must be matched with the element's tags
> 2. For deployment instance refs tags must be matched with the union of the instance tags and its model element's tags
> 3. For deployment node refs tags must be matched with the deployment node's tags
> 4. For element refs and deployment instance refs kind must be matched with the element's kind
> 5. For deployment node refs kind must be matched with the deployment node's kind
> 
> Do you agree with these rules?

Yes, seems good to me.


> I'm not sure in meaning of the NestedElementOfDeployedInstanceModel. It has instance and element properties. Does it describe an element defined inside the element from which the instance is created (defined)?

To cover the following case, when relationship has an endpoint referencing nested element:

this case

I think this model is not used for/by views, as deployment diagrams do not expand deployed instances. To filter such refs - I suggest same rules as for _deployment instance refs_ (2 and 4)

pavelpykhtin:
Completed the implementation. There are two things that bother me:
- To resolve a deployed instance of the element, I rely on compute() method (see [utills.ts:211](https://github.com/likec4/likec4/blob/6fb252115636d3896cf73970a1d0da203e3cd332/packages/core/src/compute-view/deployment-view/predicates/utils.ts#L211)) that expands a relation between elements into a set of connections between elements and ancestors of the elements. I'd rather use some kind of "resolveDeploymentInstance(element)" but have no idea if it's even feasible (as we can have more than one deployed instance)
- Parser rule for WhereRelationParticipantKind mixes ElementKinds and DeployementNodeKinds which will allow specifying a deployment node kind in element view. Tried to work around this with 
```
WhereRelation<isDeployment>:
  {infer WhereRelationKind} 'kind' EqOperator value=[RelationshipKind:Id] |
  {infer WhereRelationParticipantTag} participant=Participant StickyDot 'tag' EqOperator value=[Tag:TagId] |
  {infer WhereRelationParticipantKind} participant=Participant StickyDot 'kind' ( <!isDeployment> EqOperator value=[ElementKind:Id]) |
  {infer WhereRelationDeploymentParticipantKind} participant=Participant StickyDot 'kind' (<isDeployment> EqOperator value=[DeploymentNodeOrElementKind:Id])
;
```
but with no luck (WhereRelationDeploymentParticipantKind rule is completely ignored).

davydkov:
> Parser rule for WhereRelationParticipantKind mixes ElementKinds and DeployementNodeKinds which will allow specifying a deployment node kind in element view.

That's ok, as Langium suggests in docs - better to enforce with validation

PR #1442, opened 2025-01-16
Feat/1441 deployment element filter

### LINKED ISSUE #1441: Implement where predicates on elements in deployment views

pavelpykhtin:
Follow up #1430 
Closes #1441 

Hi, @davydkov 

One thing to mention is that I commented [this](https://github.com/likec4/likec4/blob/50288d53ed94e744fd57fda6deb98dc6e6274b17/packages/core/src/compute-view/memory/AbstractStageExclude.ts#L212C20-L212C56) line.
It was covered only by one [spec](https://github.com/likec4/likec4/blob/50288d53ed94e744fd57fda6deb98dc6e6274b17/packages/core/src/compute-view/deployment-view/clean-connections.spec.ts#L134):
Given the model
```
specification {
    element el
    deploymentNode node
}
model {
  el client 
  el cloud {
    el ui
    el backend{
      el api
    }
    el db
  }

  client -> cloud.ui
  cloud.ui -> cloud.backend.api
  cloud.backend.api -> cloud.db
}
deployment {
  node z1{
    node s1{
      instanceOf ui "ui (s1)"
      instanceOf api "api (s1)"
    }
    node s2{
      instanceOf ui "ui (s2)"
      instanceOf api "api (s2)"
    }
  }
}
views {
  deployment view some {
    include z1.**
    exclude -> z1.s2.api
    //exclude z1.s1.api
  }
}
```

commented line will convert 
image into 
image

This does not look correct. Both z1.s1.ui and z1.s2.ui don't have visible connections. They were added with the same predicate (z1.**). Thus, I expect they will behave identically. 
It also breaks the idea of explicit vs. implicit, where 'explicit' element is the one the user have explicitly added ('I want to see it on the diagram'). They must not be implicitly excluded.

### DIFF
--- a/apps/docs/src/content/docs/dsl/Deployment/views.mdx
+++ b/apps/docs/src/content/docs/dsl/Deployment/views.mdx
@@ -48,14 +48,13 @@ But they refer to deployment nodes and instances.
   The following features are not supported yet or do not work as expected:
  
   - `with` expressions
-  - `where` expressions on elements
   - Shared styles and predicates
   - Relationships browser, Element and Relationship Details popups (work with logical model)
 
   ```likec4
   deployment view prod {
     include *                                   // works
-    include * where tag is #next                // does not work
+    include * where tag is #next                // works (see details below)
     include * with { color: red }               // does not work
 
     include * -> *                              // works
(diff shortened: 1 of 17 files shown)


davydkov:
I apologize for the delay; this topic demands more thought and consideration. I was busy with refactoring diagrams and it was hard to switch contexts.

I’ll check it upcoming week.

davydkov (approved):
I apologize for the delay.

First, I wasn’t assured that `moveDisconnectedExplicitsToImplicits` must be skipped. My intention was to clean up the view by removing nodes that were connected with excluded elements. However, I realize now that the implementation needs to be improved.

For now, I agree that we can skip this step.

The rest looks good, can you rebase/update?

PR #1485, opened 2025-01-31
Added go-to-source button into nodes of the  relationship browser

pavelpykhtin:
Relationship browser is an awesome tool to navigate the model! It is especially useful when altering relationships between elements. But this use case does not work without the ability to navigate to these relations or elements which are endpoints of these relationships.


packages/diagram/src/overlays/relationships-browser/custom/ElementActions.tsx
```diff
@@ -35,6 +35,14 @@ export const ElementActions = (props: ElementActionsProps) => {
       },
     })
   }
+  buttons.push({
```
davydkov: Nice addition!
But should be added if only feature `enableVscode` is enabled

davydkov (changes requested):

packages/diagram/src/overlays/relationships-browser/custom/ElementActions.tsx
pavelpykhtin: Good point! Thanks!

davydkov (approved):

PR #1510, opened 2025-02-02
Feat(cli): Add --output-single-file flag to build give self-contained HTML file

rnwood:
This feature allows output to be bundled into a single static HTML file which does not require a web server (or relative files to load from a web server). So this is useful for anywhere you can attach/send just a single HTML file - like email, wikis and bug tracking systems.


davydkov (approved):
Wow, nice contribution!

davydkov:
I will update yarn.lock after merge

PR #1559, opened 2025-02-07
Added an option to set the listen address of serve and preview commands

### LINKED ISSUE #1526: Option to only serve on localhost
I want likec4 to be as local as possible.
If I start "likec4 start", it listens to the public interface. This is probably the case with other commands as well.
Is there any reason for this, or can 0.0.0.0 be configurable?

Ref: https://github.com/likec4/likec4/blob/v1.20.2/packages/likec4/src/vite/vite-dev.ts#L80

pavelpykhtin:
Default listen address is set to localhost

Closes #1526

### DIFF
--- a/apps/docs/src/content/docs/tooling/cli.mdx
+++ b/apps/docs/src/content/docs/tooling/cli.mdx
@@ -13,6 +13,7 @@ tableOfContents:
 ---
 
 import { LinkCard } from '@astrojs/starlight/components';
+import { Aside } from '@astrojs/starlight/components';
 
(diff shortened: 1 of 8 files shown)


davydkov (changes requested):
I suggest rename option to `--listen` (`-l`), or same as vite's `--host`
`interface` might be useful for something else 😄

davydkov:
Oh, don't sorry, vacation is always nice!

davydkov (approved):

PR #1584, opened 2025-02-10
feat: revamp search functionality

### LINKED ISSUE #1553: [Feature] Global search
Hi,

Search is already incredibly useful in big diagram. 
It would be great to be able to search elements in all diagram, not only the opened one.

Maybe it could be an option in the search popup.
If global search is activated, it will list all the views where the element is present.
If the user click on a view, the view will open centered on the element.

Regards,
Clément.

davydkov:
Preview Playground: https://d7181f09.likec4-playground.pages.dev/

SCR-20250210-plil-2

Closes #1553

Missing:
- [x] Focus on selected node
- [ ] [Fuzzy search](https://www.fusejs.io/)


Cyclonit:
This looks really nice! How do you choose which view to open when selecting an element on the left? It should default to the current view, if it contains the element. But what if it does not?

PR #1622, opened 2025-02-18
Added 'details' button to the 'relationship browser' and 'relationship details' overlays

### LINKED ISSUE #1452: [Feature request] Add "Open details" button to elements in the Browse Relationships view
The Browse Relationships view is a great way to discover relationships and dependencies between elements in the architecture, but I find that once I discover a relationship it feels a bit of a dead end.

For example, if you start with the Database view of the cloud sample (https://template.likec4.dev/view/aws_pg), you can see that the "Sessions" schema is used by the GraphQL and that in turn is used by the Customer Dashboard.

But at that point, it's hard to determine any more information about the Customer Dashboard other than viewing its relationships. It would be great if the "Open details" button that appears in the top right of elements in the normal views could also be added here. That would allow you to get more information about the Customer Dashboard including any properties and views that it appears in, which you could then navigate to.

pavelpykhtin:
Closes #1452

### DIFF
--- a/packages/diagram/src/overlays/Overlays.tsx
+++ b/packages/diagram/src/overlays/Overlays.tsx
@@ -88,28 +88,25 @@ export const Overlays = memo(() => {
 
   const onClose = useCallbackRef(() => {
     diagram.closeOverlay()
-    // relationshipsBrowserActor?.send({ type: 'close' })
-    // relationshipDetailsActor?.send({ type: 'close' })
-    // send({ type: 'close.overlay' })
   })
 
   return (
     <DiagramFeatures.Overlays>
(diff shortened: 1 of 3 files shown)


pavelpykhtin:
In [Overlays.tsx](https://github.com/likec4/likec4/pull/1622/files#diff-61a7004049f0f866444b4c0f3c4a733e80e3fe41d22029a59b8b7a5363f88b5c) onExitComplete handler is executed when active overlay and current actor has changed to other overlay (e.g. relationshipBrowser -> elementDetails). diagram.closeOverlay() closes just opened overlay.

PR #1629, opened 2025-02-20
ELK Playground - Invite to experiment

davydkov:
This PR is prepared playground and invite for you to experiment with ELK.

Check out discussion https://github.com/likec4/likec4/discussions/1591

Deployed preview:
- https://3fe14e86.likec4-playground.pages.dev/w/bigbank/elk/index/
(if you remove `/elk/` from the path you see current layout)

Experiment idea:
- React component expects DiagramView
  So we take already layouted DiagramView and layout again with ELK (just for to make an experiment simpler)
  We can use initial layout as dimensions references for leaf nodes (ignore for compounds - we get it from ELK)

- Current edges use `points` to draw bezier or `controlPoints` to draw [catmull curve](https://d3js.org/d3-shape/curve#curveCatmullRomOpen) (if edge was changed in editor).
  Easiest would be to mimic and provide `controlPoints` 
  - Maybe you get better results if align curve here with ELK (`controlPoints` is a source for the curve). (https://github.com/likec4/likec4/blob/f57c0fd1b08fee982d48f32fa490a8525366492a/packages/diagram/src/likec4diagram/custom/edges/RelationshipEdge.tsx#L23-L28


fadil-khodabuccus-cko:
Hi,

The arrows are not being rendered properly when using ELK.

image

Fadil

almereyda:
Is there a way how I could test this outside of the playground with an existing project?

Due to a high number of components, we're maintaining large graphs with likec4 and ELK promises a more condensed layout, which I would like to try and feed back experiences.

Issue #1480, opened 2025-01-30
Support global styles in relationship browser


pavelpykhtin:
Relationship browser is used in the context of the current view. Thus, it is expected that styling of elements in relationships browser will be the same as in the current view.

![Image](https://github.com/user-attachments/assets/0263e03e-d295-4d92-b354-61ed33cb293d)
```likec4
specification {
    element el

    tag red
    tag green
    tag custom
}
global {
    styleGroup elementColors {
        style element.tag = #red {
            color red
        }
        style element.tag = #green {
            color green
        }
    }
}
model {
    el sys1 {
        #red
    }
    el sys2 {
        #green
    }
    el sys3

    sys1 -> sys2 "alpha"
    sys1 -> sys2 "beta"
    sys2 -> sys3 "alpha"
    sys1 -> sys3 "alpha"
}
views {
    view index {
        global style elementColors
        include *
    }
}
```

As a solution, I suggest passing style predicates and global style includes into the relationship browser.

pavelpykhtin:
Hi, @davydkov !

I've managed to implement this for vscode preview, but the solution does not look feasible for the exported diagrams. I would be appreciated for some guidance.

My idea was based on a concept that Relationships Browser is a contextual view always (at least now) being opened from the parent view. Thus parent view should provide context for the Relationship Browser. Current view already available from the diagram context via model.view(diagram.currentView.id). Thus, it only requires extending ComputedView with 'rules' property to make styles from the parent diagram available in Relationship Browser.
But in exported diagrams the Layouted model is used and DiagramView is returned from the context. It was already quite iffy to pass rules into ComputedView, and it feels wrong to pass it to the DiagramView.

Maybe you can share some ideas on how to share styles between current view and Relationship Browser?

antoineturostowski:
hi @davydkov !
Any news about that ?
Thanks you !

Issue #1583, opened 2025-02-10
Can't filter relationships by "view" anymore ?


Kiiv:
Hi !

I think I remember that it was possible to filter the relationships in the "Relationship of" popup : only relationships visible in the current view or global. It was a dream or the feature has been removed ? :)

![Image](https://github.com/user-attachments/assets/fac8a5d8-d445-4b4d-9051-9db1b24ffc78)

Regards

davydkov:
Indeed - scope (view/global) not yet migrated

Issue #1365, opened 2024-12-18
[Feature request] Be able to export "relationships overlay" diagram


Kiiv:
Relationship overlay view is a great way to visualize relationship of a container and its internal components without needing to declare a specific "view of" the container.

It should be great to be able to export directly the diagram on this view.

PR #1667, opened 2025-02-27
Apply styles from the current view in the relationships browser

pavelpykhtin:
#1480

### DIFF
--- a/packages/core/src/compute-view/deployment-view/compute.ts
+++ b/packages/core/src/compute-view/deployment-view/compute.ts
@@ -70,6 +70,7 @@ export function computeDeploymentView<M extends AnyAux>(
 
   return calcViewLayoutHash({
     ...view,
+    rules,
     autoLayout: {
       direction: autoLayoutRule?.direction ?? 'TB',
       ...(autoLayoutRule?.nodeSep && { nodeSep: autoLayoutRule.nodeSep }),
--- a/packages/core/src/compute-view/dynamic-view/compute.ts
+++ b/packages/core/src/compute-view/dynamic-view/compute.ts
@@ -8,22 +8,22 @@ import {
   type Color,
   type ComputedDynamicView,
   type ComputedEdge,
-  DefaultArrowType,
-  DefaultLineStyle,
-  DefaultRelationshipColor,
   type DynamicView,
   type DynamicViewStep,
-  isDynamicViewParallelSteps,
-  isViewRuleAutoLayout,
-  isViewRulePredicate,
   type NonEmptyArray,
   type RelationId,
   type RelationshipArrowType,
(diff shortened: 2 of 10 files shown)


davydkov:
@pavelpykhtin, I don’t think returning and applying style rules in the UI is a great idea. You may need take into account global styles, besides it just spreads the same logic (and we will have to maintain both)

Here’s what I did in the previous version:
- First, check if the view contains the same element.
  - If it does, inherit its styles.
- If not, go up the hierarchy to find an ancestor on a view
  - If found, inherit styles from ancestor.

(Working in most cases, as relations browser "expands" connected elements)

Honestly, even though I wrote it, I already find that version tricky to follow. But this is how it worked:

https://github.com/likec4/likec4/blob/721e38bf1ff048ae36850504ff2508b348c37eff/packages/diagram/src/overlays/relationships-of/use-layouted-relationships.ts#L156-L190

Btw, maybe combine with #1583 (`view`/`global` scope)

pavelpykhtin:
@davydkov , agree this must not be done on UI level. But what if styling rules will be available on the model level?

> You may need take into account global styles, besides it just spreads the same logic (and we will have to maintain both)
Implementation in this pull request reuse methods from element/dynamic/deployment views to calculate and apply styles. Thus it follows the same pipeline as the non-overlay views. With the exception, that overlay views don't have theirs own style rules thus they borrow them from the parent view. This allows to handle global style rules and view level rules with ease.

> Here’s what I did in the previous version:

This does not work well with global rules. The whole idea of global rules is that they are the same for all views. But the suggested approach works only for elements presented on the parent view.

davydkov:
@davydkov yes, valid points.
I suggest to address them within #1365 (you probably noticed my [initial work here](packages/core/src/compute-view/relationships-view))
With this PR, as its title says, we can revise previous algorithm, but more or less keep same idea
Can you check if [`LikeC4Model.globals()`](https://github.com/likec4/likec4/blob/main/packages/core/src/model/LikeC4Model.ts#L346) is useful here?

pavelpykhtin:
Sorry, for slow replies. Busy month...

> I suggest to address them within https://github.com/likec4/likec4/issues/1365 (you probably noticed my [initial work here](https://github.com/likec4/likec4/pull/packages/core/src/compute-view/relationships-view))

Thank you! I'll take a look.

> Can you check if [LikeC4Model.globals()](https://github.com/likec4/likec4/blob/main/packages/core/src/model/LikeC4Model.ts#L346) is useful here?

This will give us definitions of global styles. But which of them should be applied? It could be a section in specification which will specify rules to apply to overlays. Something like this:
```
specification {
  ....
  overlay styles {
    global style ...
  }
}
```

Issue #1439, opened 2025-01-15
[images] File-system based icons


kieronlanning:
Is it possible to use icons/ images from the filesystem?

e.g.

```
model {
  the_system = system {
    // Relative to the .likec4 file
    icon ./images/hosting/specific-image.svg
  }
}

```

Currently I run a local Node server as part of my project just for loading images from an assets folder:

```
model {
  the_system = system {
    icon http://localhost:3010/hosting/specific-image.svg
  }
}
```

Cyclonit:
I don't see any reason for why this should not be possible. I'd give implementing this a shot.

kieronlanning:
Amazing, thanks @Cyclonit!

jentor:
As a suggestion - have an option to reuse local icons with a namespace like `tech:`.

**Example**
```zig
specification {
  iconSet 'local' './icons'

  element redpanda-topic {
    style {
      icon local:redpanda
    }
  }
}
```

kieronlanning:
While trying with the static site builder when using icons hosted on a local http server (in leu of local file system support), I've been trying a few different approaches - like replacing 'http://localhost:3020/mail.svg' with the relative/ absolute path to the file in the .likec4 file.

For example:

```dsl
# This
icon http://localhost:3020/mail.svg
# Becomes this (absolute):
icon p:/like-c4-project/assets/mail.svg
# or this, relative:
icon ../assets/mail.svg (where the likec4 file is a diagrams folder)
```

None of these approaches work, because the parsing fails with:

```
Expecting: one of these possible Token sequences:
  1. [LIB_ICON]
  2. [none]
  3. [URI_WITH_SCHEMA]
  4. [URI_RELATIVE]
```

I've also tried a post-build step of replacing the http://localhost:3020/mail.svg with /assets/mail.svg (and variations of it) in the final `.js` files, and then copying all the of referenced images into the ./assets folder in the final `./dist` folder (result of likec4 build command).

However, while the images are servable (verified by navigating to it directly), they don't show up which is strange.

If I run the original image server at the same time as the unmodified static site build the images do appear.

So, both my pre- and post- build attempts to get the images embedded in the final `./dist` folder aren't working.

Does anyone else have any thoughts or ideas I can test?

PR #1776, opened 2025-03-28
feat: added file-based icon support, as-per [images] File-system base…

kieronlanning:
Trying to add support for file-based icons, as-per #1439.

I tried running the tests/ build locally but nothing was working - build and test failures across Debian, Windows, and using the DevContainer, so this couldn't be validated I'm afraid.

I believe the Langium grammar is ok, as the `Uri` property supports what I needed to make that work:

- [Uri](https://github.com/likec4/likec4/blob/08b9882edc1eeb65fc475dde81e4cad82e614a54/packages/language-server/src/like-c4.langium#L851)
- [URI_WITH_SCHEMA](https://github.com/likec4/likec4/blob/08b9882edc1eeb65fc475dde81e4cad82e614a54/packages/language-server/src/like-c4.langium#L902) for `file://`
- [URI_RELATIVE](https://github.com/likec4/likec4/blob/08b9882edc1eeb65fc475dde81e4cad82e614a54/packages/language-server/src/like-c4.langium#L903) for `../` and `./`

Leaving this here as more of a guide in-case anyone else can get those bits running to validate/ test.

// @davydkov


kieronlanning:
Thinking about this, it wouldn't work for files at the same level as the LikeC4 file.

For example: `icon ./assets/an-image.svg`

To get that working it would require a change to the Langium grammer.

kieronlanning:
I've managed to get most of this running through WSL, but the tests are timing out which is causing issues.

I'm getting:

```bash
19) [chromium] › tests/deploy_1-gen.spec.ts:4:1 › deploy_1 - compare snapshots ───────────────────

    Test timeout of 10000ms exceeded.

    Error: locator.waitFor: Test timeout of 10000ms exceeded.
    Call log:
      - waiting for getByRole('presentation') to be visible


       6 |   await page.goto('/export/deploy_1?padding=20');
       7 |   const diagramElement = page.getByRole('presentation')
    >  8 |   await diagramElement.waitFor()
         |                        ^
       9 |   await expect(diagramElement).toHaveScreenshot('deploy_1.png', {
      10 |     omitBackground: true
      11 |   });
        at /home/b33rdy/projects/github/kieronlanning/likec4/e2e/tests/deploy_1-gen.spec.ts:8:24

    attachment #2: trace (application/zip) ─────────────────────────────────────────────────────────
    test-results/deploy_1-gen-deploy-1---compare-snapshots-chromium/trace.zip
    Usage:

        pnpm exec playwright show-trace test-results/deploy_1-gen-deploy-1---compare-snapshots-chromium/trace.zip
```

Also running into issues using playwright to run that command.

davydkov:
Hey @kieronlanning,

I believe this PR is missing a crucial aspect - it doesn’t bundle file-based icons. For instance, if I build and deploy a static website, the icons aren’t included (they’re local to you, which is why `file://…` works for you).

You’ve correctly identified the `packages/likec4/src/vite-plugin/virtuals/icons.ts`.  
I suggest that in the `code` function, you generate component for file-based icons (see [vite docs](https://vite.dev/guide/assets.html#explicit-inline-handling)):

```tsx
// Inline as base64
import Icon1 from 'path-to-icon?inline';

export const Icons = {
  // Define an icon that renders the base64
  [path-to-icon]: () => <img src={Icon1}/>;
};
```

What do you think? Can you try this? (btw, just to check, you use `pnpm dev` in `packages/likec4`, right?)

kieronlanning:
Thanks for the advice, @davydkov. Very much appreciated! I'm really keen to get this working so I don't have to spin up a separate web server for images.

I'm not a react/ fe dev, so I'm figuring this out as I go. So thanks for baring with me!

I'll take another crack at it, but I'll be without laptop for at least two weeks in the next few days, so it might have to be when I get back.

davydkov:
@kieronlanning, ping me https://discord.gg/86ZSpjKAdA or telegram [@denchieek](https://t.me/denchieek), I’m more than happy to help you.

kieronlanning:
Amazing, thanks. Just joined the Discord - I'll direct something into the support channel when I'm back.

packages/likec4/src/vite-plugin/virtuals/icons.ts
```diff
@@ -1,7 +1,7 @@
 import { filter, isString, isTruthy, pipe, unique } from 'remeda'
 import k from 'tinyrainbow'
 import { joinURL } from 'ufo'
-import type { ComputedView } from '../../model'
```
kieronlanning: I think the formatting must have removed that automatically.

packages/likec4/src/vite-plugin/virtuals/icons.ts
```diff
@@ -10,8 +10,7 @@ function code(views: ComputedView[]) {
     filter(isString),
     filter(s =>
       isTruthy(s) &&
-      !(s.toLowerCase().startsWith('http:') || s.toLowerCase().startsWith('https:') ||
-        s.toLowerCase().startsWith('file:') || s.toLowerCase().startsWith('.'))
```
kieronlanning: Dropped support for file: as it's a security risk, and blocked by most browsers.

packages/likec4/src/vite-plugin/virtuals/icons.ts
```diff
@@ -8,18 +8,33 @@ function code(views: ComputedView[]) {
...
     unique(),
   ).sort()
 
   const {
     imports,
     cases,
   } = icons.reduce((acc, s, i) => {
-    const [group, icon] = s.split(':') as ['aws' | 'azure' | 'gcp' | 'tech', string]
-
+    const isLocalImage = s.toLowerCase().startsWith('.')
```
davydkov: Since now local icons always have `file://` scheme, so

```suggestion
    const isLocalImage = s.startsWith('file:')
```

packages/likec4/src/vite-plugin/virtuals/icons.ts
```diff
@@ -8,18 +8,33 @@ function code(views: ComputedView[]) {
...
     const Component = 'Icon' + i.toString().padStart(2, '0')
 
+    if (isLocalImage) {
+      // Here the path can't be found, presumably because the file
+      // is not in the same directory as the code that's running.
+      // We need to find the path of the original file that contained
+      // the model, and then make the path (s) relative to that.
+      acc.imports.push(`import ${Component} from '${s}?inline'`)
```
davydkov, line 30: I'm not sure if we can use `file:///test/workspace/src/icon2.png` here - we need to try
Another options:
- `${withoutProtocol(s)}?inline`
  `withoutProtocol` from [ufo](https://github.com/unjs/ufo))
- `/@fs${withoutProtocol(s)}?inline` 
  Vite is doing the same, prefix with `/@fs`, but not sure if it works with `?inline`
- Calc relative path to icon from vite root (I do not pass it to virtual modules, but can be added easily)

davydkov (commented):
Hey @kieronlanning 
I think we are moving in a right direction

Based on your comment, I changed parser in [1806](https://github.com/likec4/likec4/pull/1806)

Now, all local icons have `file://` scheme and calc as relative to the document

packages/likec4/src/vite-plugin/virtuals/icons.ts
kieronlanning: Yeah, so it looks like I'll need that path so we can make it relative. The `/@fs` didn't work, or at least didn't show any affect.

The error now is exactly as your suspected, couldn't find the path.

```bash
[plugin:vite:import-analysis] Failed to resolve import "images/data-services.svg?inline" from "likec4:plugin/likec4-dev/icons.js". Does the file exist?
likec4:plugin/likec4-dev/icons.js:7:19
23 |  import Icon02 from '/workspace/packages/likec4/dev/images/report.svg?inline';
24 |  import Icon03 from '/workspace/packages/likec4/dev/images/subscription-report.svg?inline';
25 |  import Icon04 from 'images/data-services.svg?inline';
   |                      ^
26 |  import Icon05 from 'likec4/icons/tech/graphql';
27 |  import Icon06 from 'likec4/icons/tech/nodejs';
```

kieronlanning: Actually, I've removed the `file://` from the `model.c4` file, and the errors have stopped - unlikely anyone would use that anyway?

Having said that, no errors but I can't actually see any rendered models in the loaded site. It's blank paging, which is better than the errors I was seeing...

How do I validate this, @davydkov?

packages/diagram/src/context/IconRenderer.tsx
```diff
@@ -53,7 +53,10 @@ export function IconRenderer({
     return null
   }
   let icon: ReactNode
-  if (element.icon.startsWith('http://') || element.icon.startsWith('https://')) {
+  if (
+    element.icon.startsWith('http://') || element.icon.startsWith('https://') || element.icon.startsWith('../') ||
+    element.icon.startsWith('./') || element.icon.startsWith('file://')
+  ) {
```
davydkov:
```suggestion
  if (element.icon.startsWith('http://') || element.icon.startsWith('https://')) {
```

I think is the cause of your issue

packages/likec4/src/vite-plugin/virtuals/icons.ts
```diff
@@ -8,18 +8,33 @@ function code(views: ComputedView[]) {
...
+    const isLocalImage = s.startsWith('file:')
     const Component = 'Icon' + i.toString().padStart(2, '0')
 
+    if (isLocalImage) {
+      // Here the path can't be found, presumably because the file
+      // is not in the same directory as the code that's running.
+      // We need to find the path of the original file that contained
+      // the model, and then make the path (s) relative to that.
+      acc.imports.push(`import ${Component} from '${s}?inline'`)
+      acc.cases.push(`  '[${s}]': () => <img src={${Component}} />`)
```
davydkov:
```suggestion
      acc.cases.push(`  '${s}': () => jsx('img', { src: ${Component}})`)
```

`[${s}]` is invalid.

Could you try this? (Also, seems we can't use jsx syntax)

davydkov (changes requested):
I checked-out your branch, applied these two changes, and have this

image

Can you validate?

kieronlanning:
@davydkov well played, those changes seem to work!

I've been trying to get the tests to pass, specially the e2e, but they're failing in an odd way...

Complaints about JSX not being enabled.

davydkov:
@kieronlanning hmm, maybe because you push `e2e/pnpm-lock.yaml`? Lets remove it

kieronlanning:
@davydkov that's an odd one, when I run `pnpm test:e2e` from the root folder that file gets created.

Either way, still no luck. 

Would you be available to hope on a call at somepoint, try and work the problem?

davydkov:
> Would you be available to hope on a call at somepoint, try and work the problem?

Yes, around 16-18 CET

kieronlanning:
Amazing, I'm an hour behind you. Ping me on Discord when you're free...?

e2e/src/likec4/model.c4
```diff
@@ -28,18 +28,22 @@ model {
         The legacy version of our SaaS
         MVP as was presented to the first customers
       '
+      icon ./images/cloud.svg
       link ./.github/workflows/update-diagrams.yml#L19-L25 'L19-L25'
     }
 
     next = container 'Cloud Next' {
       description 'Cloud Next is the next version of our cloud systems'
+      icon file://images/cloud-next.svg
```
davydkov:
```suggestion
      // icon file://images/cloud-next.svg
```

This could be the cause.
Lets try without it?

kieronlanning:
Removed as branch has moved to this repo.

Issue #1877, opened 2025-04-25
Optionally show tags within the main visualisation of an element


bentimms:
It would sometimes be useful to see the tags on elements without needing to click the "details" button.

One solution would be to render the tags as small lozenges in one of the corners of the element.

Further ideas:
- these could be clickable to dynamically apply a filter to the view for that tag
- the view definition could specify whether all or some of the tags are shown
- the lozenge colour could be configurable in the specification of the tag

davydkov:
Actually, I had this idea of adding colors to tags in the specification, and decorate them in IDE.

bentimms:
I guess a rounded rectangle might be a more common way to display tags like this.

c4pQ:
Double the request to apply tag as a filter dynamically to the view. Found It useful while working with structurizr.

davydkov:
Released in [v1.32](https://github.com/likec4/likec4/releases/tag/v1.32.0)

PR #1988, opened 2025-06-05
fix(lsp): Fixed lsp renaming for tags and relation kinds #1819

### LINKED ISSUE #1819: [VSCode] Tag renaming via LSP is inconsistent between specification and model
Hey! Kudos for making LikeC4, this is exactly the tool I started dreaming about as soon as I learned about the C4 model ❤ 

I noticed a bug when renaming a tag (F2): the new string is applied as is in the specification **and** in the model.

```c4
specification {
    element foo
    tag myTag
}

model {
    foobar = foo {
        #myTag
    }
}
```

F2 -> `myNewTag` gives the following:

```c4
specification {
    element foo
    tag myNewTag
}

model {
    foobar = foo {
        myNewTag
    }
}
```

Thus the model definition is broken since the `#` got lost in the process. The same issue arises when typing `#myNewTag`, only it’s the specification that’s broken this time.

pavelpykhtin:
Reference to tag was defined as Hash + Id. Thus, LSP replaces the whole reference with the new value.
Splitting TagId into two terminals solves the problem at a cost of more verbose grammar.
Side effect for user - suggestions on tags are shown only after the Hash is written, on the other hand it allows us to use hash as a trigger for autocompletion.

Also, I noticed that in WhereExpressions tag value was marked as 'optional' thus the following expressions were correct:
```
include * where tag is or kind is something
```
I was not able to find any reasons for 'optional tag value' and according to unit tests it is not necessary. I removed it. Probably, that was a workaround for some problem?

## Checklist

- [ ] I've updated the documentation accordingly.
- [x] My code follows existing patterns of this project and/or improves upon them.

---

Closes #1819

### DIFF
--- a/packages/language-server/src/like-c4.langium
+++ b/packages/language-server/src/like-c4.langium
@@ -91,7 +91,7 @@ Imported:
 // Model -------------------------------------
 
 Tags:
-  (Hash values+=[Tag:Id])+ ({infer Tags.prev=current} ',' (Hash values+=[Tag:Id])*)* ';'?
+  (values+=TagRef)+ ({infer Tags.prev=current} ',' (values+=TagRef)*)* ';'?
 ;
 
 Model:
@@ -448,7 +448,7 @@ RelationExprOrWhere:
 FqnExpr:
   {infer WildcardExpression} isWildcard?='*' |
   // These are not valid in FqnExpr, but we keep for backward compatibility
-  {infer ElementTagExpression} 'element.tag' IsEqual Hash tag=[Tag:Id] |
+  {infer ElementTagExpression} 'element.tag' IsEqual tag=TagRef |
   {infer ElementKindExpression} 'element.kind' IsEqual kind=[ElementKind] |
   FqnRefExpr
 ;
@@ -502,7 +502,7 @@ WhereElementNegation:
   'not' value=WhereElementExpression;
 
 WhereElement:
-  {infer WhereElementTag} 'tag' EqOperator Hash value=[Tag:Id] |
+  {infer WhereElementTag} 'tag' EqOperator value=TagRef |
   {infer WhereElementKind} 'kind' EqOperator value=[DeploymentNodeOrElementKind]
 ;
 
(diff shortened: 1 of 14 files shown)


davydkov:
> I was not able to find any reasons for 'optional tag value' and according to unit tests it is not necessary. I removed it. Probably, that was a workaround for some problem?

Yes, this allows for permissive parsing, meaning the IDE doesn’t mark an expression as incorrect while you’re writing it.

Example.
- This is without optional (your branch):  
  SCR-20250606-kwun

- And how it is on main
  SCR-20250606-kxiq

As you see, incomplete expression did not impact next predicate.

But this is arguable optimization, as it still requires validation (that I forgot to add).
I agree with your suggestion and make it non-optional

packages/language-server/src/like-c4.langium
```diff
@@ -91,7 +91,7 @@ Imported:
 // Model -------------------------------------
 
 Tags:
-  (values+=[Tag:TagId])+ ({infer Tags.prev=current} ',' (values+=[Tag:TagId])*)* ';'?
+  (Hash values+=[Tag:Id])+ ({infer Tags.prev=current} ',' (Hash values+=[Tag:Id])*)* ';'?
```
davydkov: We have repetitive pattern, what if we introduce rule, something like:

```
TagRef:
  Hash tag=[Tag:Id];

Tags:
  (values+=TagRef)+ ({infer Tags.prev=current} ',' (values+=TagRef)*)* ';'?
```

This makes it reusable, plus easier to add semantic highlighting
(And I'm working on #1877, where I experimented with [text decoration](https://kermanx.com/reactive-vscode/examples/editor-decoration/) - having separate ast node for tag reference would help a lot)

What do you think?
And maybe same for `Dot kind=[RelationshipKind:Id]`?

pavelpykhtin: Good point! Done.

Btw, I noticed that grammar allows rules like this:
```
view some {
  include * .http
  include * -[http]->
}

dynamic view other {
  A .http B
  A -[http]-> C
}
```

but the value of 'kind' property is ignored. I guess it was copied from model expressions. If it is not used, I could clean up  it in another PR. Or it could be used as a part of definition (in a context of dynamic step) or restriction (in a context of 'include' rule).

davydkov:
> but the value of 'kind' property is ignored. I guess it was copied from model expressions. If it is not used, I could clean up it in another PR. Or it could be used as a part of definition (in a context of dynamic step) or restriction (in a context of 'include' rule).

@pavelpykhtin 
Nice, I did not notice that.
I think, under the hood, we can transform `a .uses b` to

```
a -> b where kind is uses
```

Would be great if you can do this, but I suggest after I merge types refactor (probably tomorrow)

PR #2011, opened 2025-06-16
Fixed handling of internal quotes on fromatting

pavelpykhtin:
## Checklist

- [x] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x] I've rebased my branch onto `main` **before** creating this PR.
- [x] My commit messages follow [conventional spec](https://www.conventionalcommits.org/en/)
- [x] I've added tests to cover my changes (if applicable).
- [x] I've verified that all new and existing tests have passed locally for mobile, tablet, and desktop screen sizes.
- [ ] My change requires documentation updates.
- [ ] I've updated the documentation accordingly.

Closes #2005

PR #2013, opened 2025-06-16
ci: add Windows check

davydkov:
Github check that runs build and tests on Windows
It fails on checkout main branch (expected due to `aux` filename, which is reserved on Windows)

I rebased on `fix/2005-keep-internal-quotes`, and checkout passes.

Some tests are failing, and it seems to me due to end of line character.
(if so - shall we configure `.gitattributes` or vitest snapshots to use correct one?)

I need help with this


pavelpykhtin:
Hi, @davydkov !

I've checked the failing tests. It seems that they can not be fixed with hust .gitattributes.

Two tests are related to generate-likec4-model.ts. This function combines output of generateAux() method (which is a plain string with LF symbols) and 

**generateLikeC4Model** method calls **CompositeGeneratorNode.appendTemplate** to create GenerateNode from template and **aux** substitution. **aux** variable contains a string with LF line endings. **appendTemplate** splits the provided template into separate lines and then join them together with the **NL** node. **NL** constant is defined as an instance of **NewLineNode** created without a parameters. Parameterless constructor of **NewLineNode** uses **EOL** constant as a value for the lineDelimiter, which is platform dependant:
```Typescript
// langium/src/generate/generator-node.ts:13
export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform === 'win32') ? '\r\n' : '\n';
```
Thus on Windows generateLikeC4Model returns a string with mixed line endings:
- CRLF from appendTemplate (NL implementation for windows)
- LF inside the aux substitution

langium does not provide API to globally set prefered line ending. There are two options:
- use out.append(`...`) instead of out.appendTemplate``. append method adds provided string "as is" and keeps existing line endings. But it also does not trim empty lines and indention inside the template.
- remove CR symbols from the resulting string
```Typescript
// packages\generators\src\model\generate-likec4-model.ts:31
return toString(out).replaceAll(/\r/g, '')
```


The third test is src/LikeC4.spec.ts > LikeC4 > should throw error if invalid. It fails because of path separators and trailing dot being added in Windows implementation
```
  Invalid model:
-   /workspace/source.likec4:5 Could not resolve reference to ElementKind named 'user'
+   \workspace\source.likec4:5 Could not resolve reference to ElementKind named 'user'.
```
I see two options here:
- normalize path separators and trim trailing dot before matching result
```Typescript
    const normalizeError = (e: Error) => {
      throw new Error(e.message
        .replaceAll(/\r/g, '')
        .replaceAll(/\\/g, '/')
        .replace(/\.$/g, ''))
    }

    await expect(promise.catch(normalizeError)).rejects.toThrow(`Invalid model:
  /workspace/source.likec4:5 Could not resolve reference to ElementKind named 'user'`)
```
- use regex in assert expression

I'd prefer the first option because it allows to show diff between result and expectation which does not work with .toThrow(expected: RegEx)

davydkov:
Hey @pavelpykhtin, thank you for clarifications

> 
> There are two options:
> 
> * use out.append(`...`) instead of out.appendTemplate``. append method adds provided string "as is" and keeps existing line endings. But it also does not trim empty lines and indention inside the template.
> * remove CR symbols from the resulting string
> 

I lean toward option 2 — if it leaves trimming and indention behavior untouched.  
My only concern if it's good for Windows. Any edge cases we’d hit across environments?
What do you recommend?

Also, if the generated code is committed to repositories, is this `autocrlf` intended to handle?

> * use regex in assert expression
> 
> I'd prefer the first option because it allows to show diff between result and expectation which does not work with .toThrow(expected: RegEx)

I tried this one:

```ts
    await expect(promise).rejects.toThrow(
      /source.likec4:5 Could not resolve reference to ElementKind named 'user'/
    )
```
Then I tweaked it: `source.likec4:5` -> `source...likec4:1115` and `'user'` -> `'User'`.
Still got clear, informative diffs. If on Windows behaves the same, I’d say it’s a solid tradeoff 👍🏻

pavelpykhtin:
> My only concern if it's good for Windows. Any edge cases we’d hit across environments?
What do you recommend?
Also, if the generated code is committed to repositories, is this autocrlf intended to handle?

I also prefer this option. This should not cause any problems. The result will be a file with LF only. This matches with the behavior of the git option "core.autocrlf=input" - automatically convert line endings to Unix style. This 'input' is the recommended for the crossplatform projects and well tested.


> Then I tweaked it: source.likec4:5 -> source...likec4:1115 and 'user' -> 'User'.
Still got clear, informative diffs. If on Windows behaves the same, I’d say it’s a solid tradeoff 👍🏻

On Windows the error message looks like this:
image
No highlight, but this should not be a problem. I think we can go with this solution.

davydkov:
@pavelpykhtin I think it is ready to be merged to your #2011

Can you squash it if you are ok with it?

pavelpykhtin (approved):

PR #2014, opened 2025-06-16
feat(dsl): allow self-relationships

davydkov:


bdubas-c4workx:
What is the progress on this feature?

PR #2103, opened 2025-07-18
feat(model): include project data into model and display title from likec4rc in UI

### LINKED ISSUE #2097: Page title from project properties
The title attribute could also be supported in `.likec4rc`.

- https://likec4.dev/dsl/projects/

This would already be convenient for single-project builds. It seems a little logic would have to be applied to support this for multi-project builds.

pavelpykhtin:
Hi, @davydkov!

It started as a small feature but went out of control :) The original issue #2097 was just a starting point.

The actual value of this feature is to provide access to a project's properties from the model. About a month ago I got a question from our devops if it is possible to pass some parameters through a project config. It was not possible at that moment. But now it will be easier to add.

Unfortunately, I'm not that good at typescript type system. I'm pretty sure that ProjectInfo is not an aux type and should be moved to a model, but I have no idea how to implement this. I guess, that there are some other issues. I'll be appreciate for some guidence!

## Checklist

Closes #2097


davydkov (commented):
Hey @pavelpykhtin, thanks for picking this up!

This is definitely a useful feature — I’ve had similar questions in the past too, so I clearly see the value.

I do agree with your note: the PR grew a bit too much 😄

Especially changes to `Builder`, which I think we should leave untouched — it wasn’t designed with multiple projects in mind, and I’d prefer to keep it focused for now.

Regarding the `Project` type — I actually needed it myself while working on navigation. I think we can align and merge our efforts:
https://github.com/likec4/likec4/blob/10ba4f93376ca54a89b89ce71ae4914e90b5efdf/packages/core/src/types/project.ts#L3-L8

I don’t see any strong reason to have it generic
I’d suggest something like:
```ts
interface BaseLikeC4ModelData<A extends aux.Any> {
  projectId: aux.ProjectId<A>
  project: Project
  // ...
 
}
```

As for the Builder, just an idea — what if we assign the project only in the final step?

```ts
builder
  .with(
    model(
       // ..
    ),
    views(
       // ..
    ),
  )
  .build(/* HERE */ {
     projectId,
     name
  })

```

Sort of “materialize” the builder as a project model.

pavelpykhtin:
Thank you for the review!

I reverted changes in the builder, rebased on current main, replaced generic Project and ProjectInfo with LikeC4Model.

packages/likec4/app/src/components/view-page/SelectProject.tsx
```diff
@@ -2,10 +2,10 @@ import { Button, Menu, MenuDropdown, MenuItem, MenuTarget } from '@mantine/core'
 import { IconChevronDown } from '@tabler/icons-react'
 import { Link } from '@tanstack/react-router'
 import { projects } from 'likec4:projects'
-import { useCurrentProjectd } from '../../hooks'
+import { useLikeC4ModelDataAtom } from '../../context/LikeC4ModelContext'
 
 export function SelectProject() {
-  const projectId = useCurrentProjectd()
+  const model = useLikeC4ModelDataAtom()
```
davydkov: This hook will react on changes of atom's instance, not on the changes of the value this atom holds.

Just `useLikeC4Model` fits better (or you can make `useCurrentProject` as other examples in hooks.ts and reuse [here](https://github.com/likec4/likec4/pull/2103/files#diff-f0aadca22f5aeb1f4862bcfb6ab92c759d8f9ced5f9f6285f382e488f76f01a2R26))

davydkov (approved):
Minor comment, ready to merge

packages/likec4/app/src/components/view-page/SelectProject.tsx
pavelpykhtin: Good point! Fixed.

PR #2111, opened 2025-07-24
feat: title, description and links added to spec and can be inherit in model

CookiesEater:


davydkov:
Something wrong with Github

CookiesEater:
Thank you! I've updated docs and remove separate type for string in specification

davydkov (approved):

PR #2122, opened 2025-08-03
feat(dynamic-view): support MarkdownOrString for DynamicViewStep.notes

cry999:
Hi @davydkov ,

Sorry for the delay — it took me a bit longer than expected to put this together.  
Thanks again for taking the time to walk me through the implementation!

As we discussed, this PR focuses on the **backend side** of Markdown support for dynamic view step notes.  
Here's what I've done in this PR:

- Changed `notes` type in `DynamicViewStep` and `ComputedEdge` to `MarkdownOrString`
- Updated `ViewsParser` to correctly handle Markdown-style `notes`
- Adjusted `applyCustomRelationProperties` to apply Markdown notes when overriding edges
- Added tests to:
  - validate parsing of Markdown-formatted notes in DSL
  - confirm propagation of structured notes through parsed → computed → layouted steps
- Also updated `RelationshipEdge` to access `notes.txt ?? notes.md`  
  (no UI change yet, just making sure it won’t break)

Let me know:
- if this direction aligns with what you had in mind,
- whether the current test coverage is sufficient,
- and if there’s anything I should fix or improve.

I didn’t update the documentation yet — I thought it might be better to do that together with the UI changes in the second PR. Let me know if you'd prefer otherwise.

Appreciate your time and feedback!

## Checklist


packages/language-server/src/__tests__/views-dynamic.spec.ts
```diff
@@ -171,6 +171,28 @@ describe.concurrent('dynamic views', () => {
...
+    await valid(`
+      dynamic view index1 {
+        user -> system.frontend 'User uses System'
+        system.frontend -> system.backend 'frontend uses backend'
+        system.frontend <- system.backend {
+          title 'backend uses frontend'
+          notes: "
+            # Title
+            - bullet
+          "
```
davydkov: Should be triple-quoted

```suggestion
          notes: """
            # Title
            - bullet
          """
```

davydkov:
What do you if we place walkthrough notes to the top-left, next to the panel.?

image



Here is example how it looks for view details

image

cry999:
@davydkov 

Thanks for the review!  

> What do you if we place walkthrough notes to the top-left, next to the panel.?

I think it’s a great idea — having the walkthrough notes consistently shown in the top-left will definitely improve readability.

One thing I’d suggest is setting a max height/width for the panel, just in case the `notes` content gets quite long.  
Otherwise, the panel might end up covering a large part of the diagram, which could be distracting — similar to what sometimes happens with long `description` fields.

A scrollable panel with a limit might be a good balance!

PR #2173, opened 2025-08-29
feat: added image alias support

kieronlanning:
* feat: implement configurable image aliases for @ addressing
* feat: enable addressing without config file using default aliases
* test: comprehensive test suite for @ addressing icon functionality
* fix: existing failing tests by updating icon path resolution logic to handle @ without slash correctly
* test: default workspace behaviour without configuration files tests
* feat: update Langium grammar to support @ addressing for icons
* test: updated tests to include new local image aliasing
* test: updated schema rules and tests
* chore: updated Azure icons to v22
* test: adding more test options
* feat: added support for nested likec4 files and image aliases
* docs: updates docs site with information about image aliases

## Checklist


packages/diagram/package.json
```diff
@@ -74,7 +74,7 @@
     "pack": "pnpm pack",
     "postpack": "cp likec4-diagram-$npm_package_version.tgz package.tgz || true",
     "test": "vitest run --no-isolate",
-    "clean": "pnpm rimraf dist lib bundle *.tgz",
+    "clean": "pnpm rimraf --glob dist lib bundle *.tgz",
```
kieronlanning, line 77: Makes this work on Windows.

apps/docs/src/content/docs/dsl/projects.mdx
```diff
@@ -85,3 +85,101 @@ At the moment, the following limitations apply:
...
+}
+```
+
+You can then use the alias in your model:
+
+```likec4
+// ./amazon.c4
+model {
+  serviceA = service {
+    icon: '@/service-a.png'
```
davydkov: Does it need to be a string?
In your tests I see it is not a string

packages/language-server/src/like-c4.langium
```diff
@@ -827,7 +827,7 @@ IconId returns string:
   LIB_ICON;
 
 Uri returns string:
-  URI_WITH_SCHEMA | URI_RELATIVE;
+  URI_WITH_SCHEMA | URI_RELATIVE | URI_ALIAS;
```
davydkov, line 830: btw, we also have `LinkProperty` with `Uri` as its value. Later on, aliases could be useful there as well.

kieronlanning: ...especially when it comes to relative docs, right! (:

apps/docs/src/content/docs/dsl/projects.mdx
kieronlanning: @davydkov Fixed! Great spot! 🥇

kieronlanning: Resolved in #2173 [148106b](https://github.com/likec4/likec4/pull/2173/commits/148106b203c40dbc07d0cb2e2c9f7969f86b7cdb)

PR #2386, opened 2025-11-21
fix: Highlighting nodes with notation

YousefSedik:
## Checklist

### DIFF
--- a/THIRD-PARTY-NOTICES.txt
+++ b/THIRD-PARTY-NOTICES.txt
@@ -1,0 +1,43 @@
+THIRD-PARTY SOFTWARE NOTICES AND INFORMATION
+
+This project includes modifications contributed by NVIDIA CORPORATION & AFFILIATES.
+
+================================================================================
+
+LikeC4
+https://github.com/likec4/likec4
+
+MIT License
+
+Copyright (c) 2023-2025 Denis Davydkov
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


packages/diagram/src/likec4diagram/state/machine.actions.ts
```diff
@@ -398,31 +399,26 @@ export const resetEdgesControlPoints = () =>
 
 export const notationsHighlight = (params?: { notation: ElementNotation; kind?: string }) =>
```
davydkov: Oops, I messed a bit with refactoring.
You are right, `notationsHighlight` action is called on `notations.highlight` event only.
To avoid confusion, lets remove `params`

packages/diagram/src/likec4diagram/state/machine.actions.ts
```diff
@@ -398,31 +399,26 @@ export const resetEdgesControlPoints = () =>
...
-      return Base.setDimmed(n, 'immediate')
+      const highlighted = node && shouldHighlight(node)
+      return Base.setDimmed(n, highlighted ? false : 'immediate')
     })
-    return {
-      xynodes,
-      xyedges: context.xyedges.map(Base.setDimmed('immediate')),
-    }
+
+    const xyedges = context.xyedges.map((edge) => Base.setDimmed(edge, 'immediate'))
```
davydkov: Curried `Base.setDimmed('immediate')` does not work?

davydkov:
Hey @YousefSedik 
Ready to merge? Or you still working on it?

packages/diagram/src/likec4diagram/state/machine.actions.ts
YousefSedik: i'm a little bit confused, what is the need of ```Base.setDimmed('immediate')``` anyway ? are they supposed to make the edges highlighted also ?

davydkov: When `dimmed` changed to `true` - there are animations applied
If we set to `'immediate'` - there are no animations

When user hovers element notation - we highlight related nodes. For that - we dim everything else

davydkov:
Failed spec fixed in main, update your branch

packages/diagram/src/likec4diagram/state/machine.actions.ts
YousefSedik: what about the edges ?, should they have any effect/ animation ?
if i used 
```
const xyedges = context.xyedges.map((edge) => {
      return Base.setDimmed(edge, true)
    })
 ```
it would like this, let me know what you think
 
[Screencast from 2025-11-25 18-48-42.webm](https://github.com/user-attachments/assets/4e0c8e54-923c-4225-86a4-9f7864ef3538)

PR #2401, opened 2025-11-25
feat: Support rank constraints via Graphviz/Dot layout

yubrshen:
## Checklist

- [x ] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x ] I've rebased my branch onto `main` **before** creating this PR.
- [x ] My commit messages follow [conventional spec](https://www.conventionalcommits.org/en/)
- [x ] I've verified that all new and existing tests have passed locally for mobile, tablet, and desktop screen sizes.
- [x ] My change requires documentation updates.
- [x ] I've updated the documentation accordingly.

Appendix: Design and implementation details

# Add Rank Constraints to LikeC4 Layouts

## Background

Rank directives in LikeC4 diagrams are now treated as first-class data that
travels from the DSL to rendering via Graphviz. 

## Goals and Non-goals

- **Goal:** Treat `rank` rules as a first-class concept across the DSL parser,
  the core compute pipeline, and the Graphviz layer so authors can rely on
  deterministic vertical/horizontal ordering.

### DSL and Parser Updates

The DSL now has an explicit grammar production for rank rules inside view
bodies (`packages/language-server/src/like-c4.langium`):

```langium
ViewRuleRank:
  '{' 'rank' Eq value=RankValue ';'?
    targets+=FqnRef (','? targets+=FqnRef)*
  ';'? '}'
;

RankValue:
  'same' | 'min' | 'max' | 'source' | 'sink';
```

Authors can place the block alongside existing predicates, and every target is
resolved through the standard `FqnRef` mechanism so scoping matches other view
rules. `packages/language-server/src/model/parser/ViewsParser.ts` translates the
parsed rule into a structured predicate (language-server spec tests now cover
all five buckets). Each rule keeps track of the desired `rank` bucket and its
targets, similar to how style rules work.

### Core Compute Pipeline

Changes in `packages/core/src/compute-view/element-view/compute.ts` and
`compute.spec.ts` center around the new `collectRankConstraints` helper:

- The `ElementViewRule` sequence is still resolved through the existing memory
  stages. After nodes and edges are finalized, we scan rank rules with
  `elementExprToPredicate` and retain only the nodes that actually made it into
  the computed view.
- The optional `ranks` array (typed via the new `ComputedRankConstraint` and
  exposed in `view-computed.ts` / `view-parsed.element.ts`) hangs off every
  `ComputedElementView`. Views without rank rules simply omit it to keep payload
  size minimal.
- A tiny debug hook (`globalThis.__LIKEC4_DEBUG_RANK__`) can emit the resolved
  node IDs while troubleshooting user projects.

### Layout / Graphviz Layer

Two layers consume the `ranks` array:

1. **DOT printer (`packages/layouts/src/graphviz/ElementViewPrinter.ts`):**
   - `applyExplicitRankBlocks` emits subgraphs with `rank=same|min|max|source|sink`
     for every constraint. We deduplicate node IDs to avoid duplicate DOT
     entries and return the number of rank blocks so the Graphviz adapter can
     expose it to tests (`likec4_rankBlocks` attribute).
   - Enabling Graphviz’ `newrank` and `clusterrank=global` when we tile large
     compounds ensures the manual tiling logic and explicit rank blocks do not
     fight each other.

2. **Graphviz adapter tests:** By feeding the computed view (now carrying ranks)
   through `GraphvizLayouter`, we expect the underlying WASM adapter to respect
   the explicit rank subgraphs.

### Dedicated Rank Fixture

Instead of splicing snippet nodes (`A`–`F`) into the shared fake model (which
caused unrelated snapshot diffs), we built
`packages/layouts/src/graphviz/__fixtures__/rank-snippet.ts`:

- The fixture defines a minimal LikeC4 model dump and reuses
  `computeView + withReadableEdges` so we exercise the same pipeline that real
  projects use.
- Because `withReadableEdges` omits internal bookkeeping fields, we manually
  reattach the `ranks` array (matching the snippet’s `same`, `source`, and `max`
  directives). This keeps the fixture deterministic.
- Any test that only cares about rank behaviour can import this computed view
  without touching the broader fake model.

## Testing Strategy

| Test | Location | Purpose |
| --- | --- | --- |
| `ElementViewPrinter` snapshot + rank assertions | `packages/layouts/src/graphviz/ElementViewPrinter.spec.ts` | Verifies that DOT output contains `likec4_rankBlocks`, `rank="source"`, and `rank="max"` blocks when a computed view carries ranks. Prevents regressions where the printer drops the information before Graphviz sees it. |
| Graphviz adapter snapshots | `packages/layouts/src/graphviz/wasm/__snapshots__/GraphvizWasmAdapter.spec.ts.snap` | Rebased after reverting snippet contamination so only the canonical fake model appears. Confirms that previously established views (`computedIndexView`, `computedAmazonView`, etc.) remain stable once rank support landed. |
| Rank-snippet auto-layout tests | `packages/layouts/src/graphviz/wasm/GraphvizWasmAdapter.spec.ts` | Four new tests iterate over `TB`, `BT`, `LR`, and `RL`. Each uses `computedRankSnippetView` (with ranks intact) and asserts that Graphviz returns all nodes `['A','B','C','D','E','F']`. This guards against the original disappearing-node bug across every direction we expose. |

We run the focused suite with
`pnpm --filter @likec4/layouts test -- src/graphviz/ElementViewPrinter.spec.ts src/graphviz/wasm/GraphvizWasmAdapter.spec.ts`
and rely on the existing Turbo/CI pipeline for the broader `pnpm vitest:ui` and
snapshot verification.

## Rationale Recap

- **Propagate ranks early:** Computing node membership before we pass the view to
  printers avoids heuristics or additional model lookups later.
- **Keep fixtures isolated:** A dedicated snippet fixture lets us iterate on the
  regression without destabilizing the rest of the suite.
- **Direction-wide coverage:** The original reproduction only failed in certain
  auto-layout orientations; locking all four directions keeps the behaviour
  uniform for users toggling the setting in the VS Code extension.

## Future Work

- Expose rank metadata in the VS Code preview so users can inspect which nodes
  participate in each constraint.
- Consider linting feedback when a `rank` rule matches zero nodes (currently we
  skip it silently, but surfacing a warning in the language server could reduce
  surprises).
- Investigate whether compound ranks should inherit from parent clusters, which
  might help more advanced C4 hierarchies. (Yu Shen [yubrshen] is not sure of this.)
  
## Example

An example to illustrate how to use the rank constraint has been added at 
./examples/rank-for-better-layout/demo-rank-for-better-layout.c4

## Documentation

The web site for LikeC4 has been updated on how to use rank constaints.

The update is at the end of 
./apps/docs/src/content/docs/dsl/Views/predicates.mdx

### DIFF
--- a/packages/core/src/compute-view/element-view/compute.ts
+++ b/packages/core/src/compute-view/element-view/compute.ts
@@ -1,4 +1,4 @@
-import { filter, findLast, forEach, map, pipe } from 'remeda'
+import { anyPass, filter, findLast, forEach, hasAtLeast, map, pipe } from 'remeda'
 import { ConnectionModel } from '../../model/connection/model/ConnectionModel'
 import { LikeC4Model } from '../../model/LikeC4Model'
 import type { RelationshipModel } from '../../model/RelationModel'
@@ -44,24 +44,6 @@ import { InOutRelationPredicate } from './predicates/relation-in-out'
 import { OutgoingExprPredicate } from './predicates/relation-out'
 import { WildcardPredicate } from './predicates/wildcard'
 import { buildNodes, NoFilter, NoWhere, toComputedEdges } from './utils'
-
-const rankDebugLogger: null | ((payload: { rank: string; nodes: NodeId[] }) => void) = (() => {
-  try {
-    if (typeof globalThis === 'undefined') {
-      return null
-    }
-    const flag = (globalThis as { __LIKEC4_DEBUG_RANK__?: unknown }).__LIKEC4_DEBUG_RANK__
-    if (!flag) {
-      return null
-    }
-    if (typeof console === 'undefined' || typeof console.debug !== 'function') {
-      return null
-    }
-    return (payload: { rank: string; nodes: NodeId[] }) => console.debug('[likec4][rank]', payload)
-  } catch {
-    return null
-  }
-})()
 
 function processElementPredicate(
   // ...args:
@@ -284,16 +266,15 @@ function collectRankConstraints<A extends AnyAux>(
     const nodesInRank = pipe(
       nodes,
       filter(isTargeted),
-      map(n => n.id)
+      map(n => n.id),
     )
-    if (!hasAtLeast(nodesInRank,2)) {
+    if (!hasAtLeast(nodesInRank, 1)) { // rank value sink, source, min, max can admit one node
       continue
     }
     constraints.push({
       type: rule.rank,
       nodes: nodesInRank,
     })
-    rankDebugLogger?.({ rank: rule.rank, nodes: nodesInRank })
   }
   return constraints
 }
--- a/packages/core/src/types/view-computed.ts
+++ b/packages/core/src/types/view-computed.ts
@@ -15,6 +15,7 @@ import type {
 } from './scalar'
 import type {
   BaseViewProperties,
+  RankValue,
   ViewAutoLayout,
   ViewManualLayout,
   ViewWithHash,
--- a/packages/language-server/src/like-c4.langium
+++ b/packages/language-server/src/like-c4.langium
@@ -348,9 +348,9 @@ ViewRule:
 ;
 
 ViewRuleRank:
-  '{' 'rank' Eq value=RankValue ';'?
-    targets+=FqnRef (','? targets+=FqnRef)*
-  ';'? '}'
+  'rank' value=RankValue? '{'
+    targets=FqnExpressions
+  '}'
 ;
 
 RankValue returns string:
--- a/packages/language-server/src/model/__tests__/model-parser.spec.ts
+++ b/packages/language-server/src/model/__tests__/model-parser.spec.ts
@@ -1010,14 +1010,15 @@ describe.concurrent('LikeC4ModelParser', () => {
           block root {
             block child {
               block grand
+              block grand2
             }
           }
         }
         views {
           view index {
             include *
-            {rank=same; root child}
-            {rank=max; root.child.grand}
+            rank same { root.child.grand, root.child.grand2 }
+            rank max { root.child.grand }
           }
         }
       `)
@@ -1030,12 +1031,12 @@ describe.concurrent('LikeC4ModelParser', () => {
           targets: [
             {
               ref: {
-                model: 'root',
+                model: 'root.child.grand',
               },
             },
             {
               ref: {
-                model: 'root.child',
+                model: 'root.child.grand2',
               },
             },
           ],
--- a/packages/language-server/src/validation/view-checks.ts
+++ b/packages/language-server/src/validation/view-checks.ts
@@ -3,23 +3,40 @@ import { ast } from '../ast'
 import type { LikeC4Services } from '../module'
 import { tryOrLog } from './_shared'
 
-export const viewRuleRankChecks = (services: LikeC4Services): ValidationCheck<ast.ViewRuleRank> => {
+// Helper to collect FqnExpr values from FqnExpressions linked list
+function collectFqnExprs(exprs: ast.FqnExpressions | undefined): ast.FqnExpr[] {
+  const result: ast.FqnExpr[] = []
+  let iter: ast.FqnExpressions | undefined = exprs
+  while (iter) {
+    if (iter.value) {
+      result.push(iter.value)
+    }
+    iter = iter.prev
+  }
+  return result.reverse()
+}
+
+export const viewRuleRankChecks = (_services: LikeC4Services): ValidationCheck<ast.ViewRuleRank> => {
   return tryOrLog((el, accept) => {
-    if (el.targets.length < 2 && el.value === 'same') {
+    const targetExprs = collectFqnExprs(el.targets)
+
+    if (targetExprs.length < 2 && el.value === 'same') {
       accept('warning', 'Rank rule should have at least 2 targets', {
         node: el,
         property: 'targets',
       })
     }
 
-    const firstParent = el.targets[0]?.parent
-    for (let i = 1; i < el.targets.length; i++) {
-      const target = el.targets[i]
-      if (!areSame(firstParent, target?.parent)) {
-        accept('error', 'All targets must have the same parent', {
+    // Filter to only FqnRefExpr for parent comparison
+    const fqnRefExprs = targetExprs.filter(ast.isFqnRefExpr)
+
+    const firstParent = fqnRefExprs[0]?.ref?.parent
+    for (let i = 1; i < fqnRefExprs.length; i++) {
+      const target = fqnRefExprs[i]
+      if (el.value === 'same' && !areSame(firstParent, target?.ref?.parent)) {
+        accept('error', 'All targets must have the same parent rank same', {
           node: el,
           property: 'targets',
-          index: i,
         })
       }
     }
--- a/apps/docs/src/content/docs/dsl/Views/predicates.mdx
+++ b/apps/docs/src/content/docs/dsl/Views/predicates.mdx
@@ -843,28 +843,24 @@ views {
 You can keep specific elements on the same horizontal/vertical level (or push them to the beginning/end of the layout)
 by adding an explicit `rank` block to the view.
 These rank constraints are forwarded to the Graphviz layout engine to achieve the desired layout effects.
-(Note, the syntax of the rank constraints is the same as that of Graphviz with Dot layout engine.)
 
 ```likec4
 view checkoutFlow {
   include *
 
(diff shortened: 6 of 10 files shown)


packages/core/src/compute-view/element-view/compute.ts
```diff
@@ -41,6 +45,24 @@ import { OutgoingExprPredicate } from './predicates/relation-out'
...
+      return null
+    }
+    if (typeof console === 'undefined' || typeof console.debug !== 'function') {
+      return null
+    }
+    return (payload: { rank: string; nodes: NodeId[] }) => console.debug('[likec4][rank]', payload)
+  } catch {
+    return null
+  }
+})()
```
davydkov: This seems was used during development only.
Lets take this out for now:
- Needs to be more "centralized" or being well defined "common" approach for logging in core (though, agree, there is none yet)
- Scoped to ranks only (even the name is `__LIKEC4_DEBUG_RANK__`, not a `__LIKEC4_DEBUG__`).
- Expects _"feature flag"_ on `globalThis`, but no instructions how to enable it.  
 (prefer `esm-env` or [import maps](https://nodejs.org/api/packages.html#subpath-imports))
- Maybe we should just add `@likec4/log` to dependencies and use "trace" level

packages/core/src/compute-view/element-view/compute.ts
```diff
@@ -244,9 +267,39 @@ export function computeElementView<A extends AnyAux>(
...
+    const seen = new Set<NodeId>()
+    for (const node of nodes) {
+      if (predicates.some(predicate => predicate(node))) {
+        seen.add(node.id)
+      }
+    }
+    if (seen.size === 0) {
+      continue
+    }
+    const nodesInRank = [...seen]
```
davydkov: We can rewrite with remeda 

```suggestion
    const isTargeted = anyPass(rule.targets.map(elementExprToPredicate))
    const nodesInRank = pipe(
      nodes,
      filter(isTargeted),
      map(n => n.id)
    )
    if (!hasAtLeast(nodesInRank,2)) {
      continue
    }
```


Then `ComputedRankConstraint`:

```ts
export interface ComputedRankConstraint {
  type: 'same' | 'min' | 'max' | 'source' | 'sink'
  /**
   * At least two nodes to be ranked
   */
  nodes: [NodeId, NodeId, ...NodeId[]]
}
```

packages/core/src/types/view-computed.ts
```diff
@@ -102,6 +102,11 @@ export interface ComputedEdge<A extends AnyAux = AnyAux> extends aux.WithOptiona
   dir?: 'forward' | 'back' | 'both'
 }
 
+export interface ComputedRankConstraint {
+  type: 'same' | 'min' | 'max' | 'source' | 'sink'
```
davydkov:
```suggestion
  type: RankValue
```

packages/language-server/src/model/__tests__/model-parser.spec.ts
```diff
@@ -1000,6 +1000,59 @@ describe.concurrent('LikeC4ModelParser', () => {
...
+              ref: {
+                model: 'root',
+              },
+            },
+            {
+              ref: {
+                model: 'root.child',
+              },
+            },
+          ],
```
davydkov, line 1042: Is it valid that parent-child can be same-ranked?

Maybe same validation check as in relationships?
https://github.com/likec4/likec4/blob/89cadcd084b7d5e5572dc335c0e1c14ef59ae217/packages/language-server/src/validation/relation.ts#L44-L48

packages/language-server/src/validation/view-checks.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const firstParent = el.targets[0]?.parent
+    for (let i = 1; i < el.targets.length; i++) {
+      const target = el.targets[i]
+      if (!areSame(firstParent, target?.parent)) {
+        accept('error', 'All targets must have the same parent', {
+          node: el,
+          property: 'targets',
+          index: i,
+        })
```
davydkov: Why do we require this? 

Can I rank two sinks from different brokers?
```
{ rank = sink;
  broker1.some-topic
  broker2.some-other-topic
}
```

packages/core/src/compute-view/element-view/compute.ts
davydkov: Oh, I see, we may have, for example, `sink` rank with just one node

davydkov:
Hey @yubrshen 
I understand you tried to keep the syntax close to Graphviz, but I think the following fits better:

```
  view index {
    title 'Effect or Rank'

    include *

    rank min {
      A B
    }

    rank { // default to same keyword,
      A B
    }
```


1. It follows the same approach as the current rules, which is better for users: they expect consistent patterns and won’t be frustrated by a special case here.  
  We keep the `Rule Block{ }` pattern (instead of `Block{ Rule }`).  
  It also plays nicer with code completion, since `rank` will be suggested at the right place. With your variant, completion only kicks in inside the block.

2. Non-ambiguous.  
  Imagine, later we introduce something like `RuleA OptionalBlock{ }`. The parser would then have to guess whether `OptionalBlock` is part of `RuleA`, or the beginning of a `ViewRuleRank`. With the proposed syntax, that confusion disappears.

packages/language-server/src/like-c4.langium
```diff
@@ -343,9 +343,19 @@ ViewRule:
...
   ViewRuleGroup |
   ViewRuleStyleOrGlobalRef |
-  ViewRuleAutoLayout
+  ViewRuleAutoLayout |
+  ViewRuleRank
+;
+
+ViewRuleRank:
+  '{' 'rank' Eq value=RankValue ';'?
+    targets+=FqnRef (','? targets+=FqnRef)*
```
davydkov: I suggest to reuse `FqnExpressions` from `ViewRuleStyle`:

https://github.com/likec4/likec4/blob/9102287131135c017ae263e91cf963fa2b0e6bc2/packages/language-server/src/like-c4.langium#L433-L434

As Langium team suggests, it’s better to enforce this via validation rather than in the parser, because then we can give a precise error message explaining why the input is invalid.

yubrshen:
> Hey @yubrshen I understand you tried to keep the syntax close to Graphviz, but I think the following fits better:
> 
> ```
>   view index {
>     title 'Effect or Rank'
> 
>     include *
> 
>     rank min {
>       A B
>     }
> 
>     rank { // default to same keyword,
>       A B
>     }
> ```
> 
> 1. It follows the same approach as the current rules, which is better for users: they expect consistent patterns and won’t be frustrated by a special case here.
>    We keep the `Rule Block{ }` pattern (instead of `Block{ Rule }`).
>    It also plays nicer with code completion, since `rank` will be suggested at the right place. With your variant, completion only kicks in inside the block.
> 2. Non-ambiguous.
>    Imagine, later we introduce something like `RuleA OptionalBlock{ }`. The parser would then have to guess whether `OptionalBlock` is part of `RuleA`, or the beginning of a `ViewRuleRank`. With the proposed syntax, that confusion disappears.

The grammar changed as 
```
ViewRuleRank:
  'rank' value=RankValue? '{'
    targets=FqnExpressions
  '}'
;
```

packages/language-server/src/model/__tests__/model-parser.spec.ts
yubrshen: Simplify this positive test case for rank same case of the same siblings. 
I have decided not support same-ranked for parent-child. Such case has been rejected in 
view-checks.ts:33-42

packages/core/src/compute-view/element-view/compute.ts
yubrshen: Removed the code and the reference to the function.

packages/language-server/src/like-c4.langium
yubrshen: Changed the grammar per suggestions including using targets=FqnExpressions

packages/language-server/src/validation/view-checks.ts
yubrshen: Only requires the same parent for rank same, no longer requiring that for the other rank value. 
It remains to observe if there is any complication for the relaxed constraint.

yubrshen:
@davydkov accepted all your feedback, completed the fixes, update with latest of the base, and passed all test. Notice that the button for "Merge pull request" is active. Please advise, if I should make the request to merge pull request now, or you'll do more review. Thanks for helping me to get this first pull request realized!

PR #2879, opened 2026-04-11
fix(core): display images in markdown descriptions

### LINKED ISSUE #2505: Images don't display in markdown
### Relates to

LikeC4 Language

### Version you are using

1.46.3

### Describe the issue

Images within markdown don't display. I'm using the markdown syntax for an image, like:
    image text

I've tried:
- local image file with relative and absolute paths
- local image in same directory as c4 files
- url to remote image

All of the above result in the same broken image icon.

_No response_

ckeller42:
## Summary

Fix for #2505 — images in markdown descriptions (`alt`) were not rendering.

### Root Cause

The `rehype-sanitize` configuration in `to-html.ts` used `defu()` to merge custom SVG-related tags with the default sanitizer schema. However, `defu` replaces arrays instead of merging them — the custom `tagNames: ['svg', 'g', 'path', 'div']` completely overwrote `defaultSchema.tagNames` (41 tags including `img`, `a`, `p`, etc.), causing all standard HTML tags to be stripped.

### Fix

Replaced `defu()` with spread-based merging that appends custom tags/attributes to the default lists instead of replacing them.

### Files changed

- `packages/core/src/utils/markdown/to-html.ts` — spread-based schema merge
- `.changeset/fix-markdown-images.md` — patch changeset for `@likec4/core`

Closes #2505

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/fix-markdown-images.md
+++ b/.changeset/fix-markdown-images.md
@@ -1,0 +1,5 @@
+---
+'@likec4/core': patch
+---
+
+Fix images not displaying in markdown descriptions by preserving default sanitizer tag list
--- a/packages/core/src/utils/markdown/to-html.ts
+++ b/packages/core/src/utils/markdown/to-html.ts
@@ -1,4 +1,3 @@
-import defu from 'defu'
 import rehypeSanitize, { defaultSchema } from 'rehype-sanitize'
 import rehypeStringify from 'rehype-stringify'
 import remarkGfm from 'remark-gfm'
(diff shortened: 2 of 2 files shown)


ckeller42:
@coderabbitai review

ckeller42:
Closing — the `defu` merge was already working correctly. The root cause is that single-quoted descriptions (`description "image"`) are stored as plain text (`{ txt: ... }`) and the `html` getter returns them without markdown processing. Only triple-quoted descriptions (`description '''image'''`) get markdown rendering. Need to investigate whether this is by design or if the `html` getter should try markdown parsing for all descriptions.

PR #2910, opened 2026-04-16
fix(vscode): include all runtime deps in extension package

ckeller42:
## Summary

Fix VSCode extension language server crash on v1.55.1 — "Cannot find package 'bundle-require'".

## Root Cause

The `package-vsix.mts` script used `pnpm ls -P --json` to discover production dependency versions. But pnpm doesn't list dependencies that are hoisted through devDependencies — `bundle-require` and `esbuild` are declared in `package.json` but resolved through `@likec4/lsp` (a devDep), so `pnpm ls -P` returned only 2 of 4 declared deps.

**Before:** VSIX included only `fdir`, `std-env`
**After:** VSIX includes `bundle-require`, `esbuild` (as esbuild-wasm), `fdir`, `load-tsconfig`, `std-env`

## Fix

Replace `pnpm ls -P --json` with `pnpm why <pkg> --json` for each declared dependency. `pnpm why` finds packages anywhere in the dependency graph regardless of hoisting.

## Files changed

| File | Change |
|---|---|
| `packages/vscode/scripts/package-vsix.mts` | Use `pnpm why` instead of `pnpm ls -P` for version resolution |

## Test plan

- [x] `pnpm package-vsix` produces VSIX with all 5 runtime deps in node_modules
- [x] Installed extension starts language server without "Cannot find package" errors
- [x] CI passes

### DIFF
--- a/.changeset/fix-vscode-packaging-deps.md
+++ b/.changeset/fix-vscode-packaging-deps.md
@@ -1,0 +1,5 @@
+---
+'likec4-vscode': patch
+---
+
+Fix VSCode extension packaging to include all runtime dependencies (bundle-require, esbuild, load-tsconfig). The language server failed to start with "Cannot find package" errors because `pnpm ls -P` didn't list deps that were hoisted through devDependencies.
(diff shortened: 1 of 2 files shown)


ckeller42:
@coderabbitai review

ckeller42:
@coderabbitai review

ckeller42:
@coderabbitai review

davydkov:
I've updated `package.json` of vscode plugin with explicit list of runtime deps - should be enough for now

PR #2968, opened 2026-05-21
feat(cli): add --public option for Vite publicDir

### LINKED ISSUE #1941: LikeC4 deletes all other files in the output folder
### Relates to

CLI

### Version you are using

1.30.0

### Describe the issue

When using the `--output-single-file` option, likec4 deletes all files in the output folder except `index.html`. I have some .png files for instance that I link to in various elements of the model but they are deleted.

It should, in my opinion, only delete the temporary files that are embedded in the html and nothing else.

_No response_

farhan523:
## Checklist

## Summary

Resolves #1941.

Adds `--public` (alias `--public-dir`) to `likec4 build` and `likec4 start` (`dev`/`serve`). Files in this directory are copied into the build output as-is via Vite's `publicDir` — and, importantly, are preserved when `--output-single-file` is used.

Today the single-file build cleans up `outDir` by deleting every file that isn't `index.html`. That step also wipes out user assets sitting in `outDir` (the original report: PNGs linked from views disappearing). The new flag lets users keep a folder of static assets that the build copies through and the single-file cleanup leaves alone.

Implementation approach matches the maintainer's hint in #1941 ("Setting the [public dir](https://vite.dev/guide/assets#the-public-directory) may solve your issue. (to be implemented in CLI)").

## Implementation

- New option in `packages/likec4/src/cli/options.ts`, wired into both `build` and `serve` commands.
- `viteBuild` / `viteDev` copy the user-provided directory's contents into the internal temp `publicDir` before invoking Vite, so all existing assets (favicon, robots.txt, webcomponent bundle) keep working unchanged.
- Extracted the single-file cleanup into `removeAllButPreserved(outDir, preserved)` so the same deletion logic now also preserves the user's top-level entries (in addition to `index.html`).
- Errors early with a clear message if the path doesn't exist or isn't a directory.

## Out of scope

This PR exposes the option as a CLI flag only, per the maintainer's note in #1941 ("to be implemented in CLI"). A `likec4.config` schema entry could be a follow-up if there's interest.

## Test plan

New unit tests (run via `pnpm test`):

- [x] `copyUserPublicDir` copies all top-level entries and returns their names (`packages/likec4/src/vite/utils.spec.ts`)
- [x] `copyUserPublicDir` returns `[]` for an empty dir
- [x] `copyUserPublicDir` throws a clear error when the path is missing / not a dir
- [x] `removeAllButPreserved` keeps `index.html` only when no user dir provided (`packages/likec4/src/vite/vite-build.spec.ts`)
- [x] `removeAllButPreserved` keeps `index.html` + user files (the actual #1941 fix)
- [x] `removeAllButPreserved` is a no-op when nothing to remove
- [x] `serve` handler propagates `userPublicDir` to `viteDev` (extends `serve.spec.ts`)

Local checks:

- [x] `pnpm test` — 11 / 11 new tests passing
- [x] `pnpm --filter likec4 typecheck` — clean
- [x] `pnpm lint:errors-only` — clean (pre-existing warnings unchanged)
- [x] `pnpm fmt` — applied

### DIFF
--- a/.changeset/cli-public-dir-option.md
+++ b/.changeset/cli-public-dir-option.md
@@ -1,0 +1,5 @@
+---
+'likec4': minor
+---
+
+Add `--public` option (alias `--public-dir`) to `likec4 build` and `likec4 start` for specifying a directory that Vite serves and copies as-is into the output (Vite's `publicDir`). Files in this directory are preserved in the build output, including when `--output-single-file` is used. Resolves #1941.
--- a/apps/docs/src/content/docs/tooling/cli.mdx
+++ b/apps/docs/src/content/docs/tooling/cli.mdx
@@ -101,13 +101,14 @@ likec4 dev
 This recursively searches for `*.c4` and `*.likec4` files in the current folder, parses them, and serves diagrams via a local web server.  
 Any change in the source files triggers a hot update in the browser immediately.
 
-| Option              | Description                                                                                               |
-| ------------------- | --------------------------------------------------------------------------------------------------------- |
-| `--port`            | Port for the dev server (default: 5173, or `PORT` env var)                                                |
-| `--hmr-port`        | Port for the HMR WebSocket (default: auto-discovered from 24678–24690, or `HMR_PORT` env var)             |
-| `--listen, -l`      | IP address to listen on (default: `127.0.0.1`, or `0.0.0.0` inside a container)                          |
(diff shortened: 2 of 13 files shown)


davydkov (approved):

davydkov:
Hey @farhan523!
Thank you for contribution

PR #2971, opened 2026-05-22
feat(cli): add --allowed-host option for dev server

### LINKED ISSUE #1650: feat: cli option `--allowed-host`
In #1649 we temporary enabled access to any host
This is not recommended and we should enable only strictly defined

(somehow env `__VITE_ADDITIONAL_SERVER_ALLOWED_HOSTS` did not work, as said in vite docs, need further investigation)

farhan523:
## Checklist

## Summary

Resolves #1650.

Adds a repeatable `--allowed-host` option to `likec4 start` (aliases `serve` / `dev`) which threads through to Vite's [`server.allowedHosts`](https://vite.dev/config/server-options#server-allowedhosts).

Today, `vite-dev.ts` hardcodes `allowedHosts: true` with a `// TODO: temprorary enable access to any host` comment from #1649. This PR keeps `true` as the default (so existing behaviour is preserved and the change is strictly additive), and gives users a CLI surface to scope access down when they want it.

```sh
likec4 start --allowed-host foo.local --allowed-host bar.example.com
```

## Implementation

- New `allowedHost` option in `packages/likec4/src/cli/options.ts` (`array: true`, `string: true`, `requiresArg: true`).
- Wired through `serve/index.ts` → `serve.ts` handler → `viteDev`.
- In `vite-dev.ts`, replaces the hardcoded `allowedHosts: true` with `allowedHosts && allowedHosts.length > 0 ? allowedHosts : true`. The obsolete `// TODO: temprorary…` comment is removed since users now have a way to address this themselves.
- Added two propagation tests to `serve.spec.ts` (with the hostnames + without — default branch).
- Added `--allowed-host` row to the `serve` options table in `apps/docs/src/content/docs/tooling/cli.mdx`.

## Out of scope

This PR does **not** change the default. The maintainer note in #1650 says *"we should enable only strictly defined"* — that's a behaviour change that probably deserves its own PR (and possibly a deprecation cycle), separate from adding the CLI surface. Happy to follow up if you'd like the default flipped here.

## Test plan

```sh
pnpm test --no-coverage --no-typecheck packages/likec4/src/cli/serve/serve.spec.ts
# Tests  6 passed (6)

pnpm --filter likec4 typecheck   # clean
pnpm lint:errors-only             # 0 errors
pnpm fmt                          # applied
```

New propagation tests:

- [x] `serve` handler propagates `allowedHosts` array to `viteDev`
- [x] `serve` handler leaves `allowedHosts` undefined when not provided (preserves the "allow all" default)

### DIFF
--- a/.changeset/cli-allowed-host-option.md
+++ b/.changeset/cli-allowed-host-option.md
@@ -1,0 +1,5 @@
+---
+'likec4': patch
+---
+
+Add `--allowed-host` option to `likec4 start` (`serve` / `dev`) for scoping which hostnames are allowed to access the dev server (Vite's `server.allowedHosts`). Can be repeated. When omitted, all hosts are allowed (current behaviour). Resolves #1650.
--- a/apps/docs/src/content/docs/tooling/cli.mdx
+++ b/apps/docs/src/content/docs/tooling/cli.mdx
@@ -109,6 +109,7 @@ Any change in the source files triggers a hot update in the browser immediately.
 | `--no-react-hmr`          | Disable React HMR (useful for CI or static previews)                                                      |
 | `--use-dot`               | Use local Graphviz (`dot`) binaries instead of bundled WASM                                               |
 | `--public, --public-dir`  | Directory whose files are served as-is (Vite [`publicDir`](https://vite.dev/guide/assets#the-public-directory)) |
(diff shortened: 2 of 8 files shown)


davydkov (approved):
@farhan523 Thanks for the PR! I agree with your "Out of scope" reasoning — let's keep `allowedHosts: true` as the default for now.

Changing it to "strictly defined" is a behaviour change that would break existing setups, so it's better keep it (for 1.x)

The additive approach here is exactly right: users who need to scope hosts down have a clean CLI surface, everyone else keeps working as-is.

PR #2105, opened 2025-07-21
feat: color of text in elements is now calculated with chroma-js and based on APCA algorithm to determine the best contrast (#2100) 

Kiiv:
## Checklist

- [x] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x] I've rebased my branch onto `main` **before** creating this PR.
- [x] My commit messages follow [conventional spec](https://www.conventionalcommits.org/en/)
- [x] I've added tests to cover my changes (if applicable).
- [x] I've verified that all new and existing tests have passed locally for mobile, tablet, and desktop screen sizes.
- [ ] My change requires documentation updates.
- [ ] I've updated the documentation accordingly.

PR #2978, opened 2026-05-23
feat(diagram): accurate text color for tags with custom colors

### LINKED ISSUE #2143: [feature request] accurate text color for tags
Users can set [custom color for tags](https://likec4.dev/dsl/specification/#tag):
```zig
specification {
  tag deprecated {
    color #FF0000
  }
}
```
This color is used for background, and we derive text color with CSS trick:
https://github.com/likec4/likec4/blob/c53f127b1fe6af81658c2eb8dc6b8e41f6162465/styled-system/preset/src/recipes/likec4tag.ts#L34-L40

This is not ideal, and @Kiiv in #2105 suggested another solution.

We should apply the same in [`TagStylesProvider`](https://github.com/likec4/likec4/blob/main/packages/diagram/src/context/TagStylesContext.tsx) and generate `--colors-likec4-tag-text` if tag has a color defined in the specification.
Plus cleanup `autoTextColor` from panda recipe

farhan523:
## Checklist

- [ ] My change requires documentation updates. _(no — internal styling change, no public DSL/CLI surface)_
- [x] I've updated the documentation accordingly. _(N/A)_

## Summary

Resolves #2143.

Tags with a custom color in the specification (e.g. `tag deprecated { color #FF0000 }`) now get an accurate, accessible text color derived from the background via the [APCA](https://github.com/Myndex/apca-w3) algorithm, instead of relying on the previous CSS-filter workaround (`autoTextColor` variant in the `likec4tag` recipe).

This is the cleanup the issue describes: emit `--colors-likec4-tag-text` from `TagStylesProvider` whenever a custom color is set, and drop the `autoTextColor` recipe variant.

## Implementation

- **`@likec4/core/styles`** — export `getContrastedColorsAPCA` (already used internally by `computeColorValues` for element colors since [#2105](https://github.com/likec4/likec4/pull/2105)). Reuses the existing chroma-js + APCA pipeline so we don't introduce a second contrast implementation.
- **`@likec4/diagram` — [`TagStylesContext.tsx`](https://github.com/likec4/likec4/blob/main/packages/diagram/src/context/TagStylesContext.tsx)** — in `generateColorVars`, the custom-color branch now computes `getContrastedColorsAPCA(color).hiContrast` and emits `--colors-likec4-tag-text` alongside `--colors-likec4-tag-bg` and `--colors-likec4-tag-bg-hover`. Radix-named tags are unchanged.
- **`@likec4/style-preset` — [`likec4tag.ts`](https://github.com/likec4/likec4/blob/main/styled-system/preset/src/recipes/likec4tag.ts)** — the `autoTextColor` variant (including the `staticCss` entry, `defaultVariants`, and CSS-filter trick) is removed. Inner-span color is set unconditionally to `likec4.tag.text` in `base`.
- **`ElementTags.tsx`** — drops the `autoTextColor: isTagColorSpecified(spec)` prop on the `likec4tag()` call. The `useTagSpecification` import is removed (no longer needed in this file).

## Test plan

New unit tests in `packages/diagram/src/context/TagStylesContext.spec.ts`:

- [x] Emits all three CSS vars (bg, bg-hover, text) for a custom hex color, with a derived text color
- [x] Derives a light text color for dark custom backgrounds (`#000000` → text channels > 0x99)
- [x] Derives a dark text color for light custom backgrounds (`#ffffff` → text channels < 0x66)
- [x] Still emits radix-based vars (including text) for a named tag color (`tomato`)
- [x] Uses `dark-2` text for radix backgrounds that are light at scale 9 (grass, lime, yellow, amber)
- [x] Returns `''` for an unrecognised color

Local checks: `pnpm test packages/diagram/src/context/TagStylesContext.spec.ts` → 6/6 passing, `pnpm --filter @likec4/core --filter @likec4/diagram --filter @likec4/style-preset typecheck` → clean, `pnpm lint:errors-only` → 0 errors.

Generated panda artifacts (`packages/react/styled-system/recipes/likec4tag.{mjs,d.mts}` and the workspace `styled-system/styles/dist/`) are git-ignored per AGENTS.md, so CI's `pnpm generate` step will produce the updated outputs.

Thanks to @Kiiv for the APCA contrast work in [#2105](https://github.com/likec4/likec4/pull/2105) — this PR reuses that infrastructure.

### DIFF
--- a/.changeset/tag-text-contrast.md
+++ b/.changeset/tag-text-contrast.md
@@ -1,0 +1,7 @@
+---
+'@likec4/core': patch
+'@likec4/diagram': patch
+'@likec4/style-preset': patch
+---
+
+Tags with custom hex / rgb colors now get an accurate text color derived from the background via APCA contrast, instead of the previous CSS-filter workaround. `TagStylesProvider` emits `--colors-likec4-tag-text` for all tags (custom-colored and named), and the `autoTextColor` variant is removed from the `likec4tag` recipe. `getContrastedColorsAPCA` is now exported from `@likec4/core/styles`. Resolves #2143.
--- a/packages/core/src/styles/compute-color-values.ts
+++ b/packages/core/src/styles/compute-color-values.ts
@@ -18,6 +18,15 @@ type ColorPalette = {
   rel_hiContrast: HexColor
 }
 
+/**
+ * Returns true when the input is a valid color literal parseable by chroma-js
+ * (e.g. `#ff0000`, `rgb(255, 0, 0)`, named CSS colors). Useful as a safety
+ * check before calling color-math helpers that throw on malformed input.
+ */
+export function isValidColor(color: string | chroma.Color): boolean {
+  return chroma.valid(color)
+}
+
 export function computeColorValues(color: ColorLiteral): ThemeColorValues {
   invariant(chroma.valid(color), `Invalid color: ${color}`)
   const normalizedRefColor = color.trim()
--- a/packages/core/src/styles/index.ts
+++ b/packages/core/src/styles/index.ts
@@ -1,5 +1,5 @@
 export { assignTagColors, DefaultTagColors } from './assignTagColors'
-export { computeColorValues, getContrastedColorsAPCA } from './compute-color-values'
+export { computeColorValues, getContrastedColorsAPCA, isValidColor } from './compute-color-values'
 export { styleDefaults } from './defaults'
 export { defaultTheme } from './theme'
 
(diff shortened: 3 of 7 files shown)


farhan523:
Pushed [56f8cb9](https://github.com/likec4/likec4/pull/2978/commits/56f8cb921) addressing the CodeRabbit finding on validation.

Verified the concern against current code:
- isTagColorSpecified is a prefix check only (accepts `#zzz`, `rgb(garbage)`, …).
- getContrastedColorsAPCA calls `chroma(refColor)` with no guard — it'd throw on malformed input and crash the diagram render. Real bug.

Fix:
- Added a tiny isValidColor helper in `@likec4/core/styles` (wraps `chroma.valid`).
- generateColorVars now requires both isTagColorSpecified *and* isValidColor before invoking APCA. Malformed values fall through to the default tag styling.
- New regression test in TagStylesContext.spec.ts covers `#zzzzzz`, `#12`, `rgb(garbage)` — `expect(() => ...).not.toThrow()` + returns `''`. All 7 tests passing.

Diff: 4 files, +25 / -3.

davydkov (approved):

PR #2984, opened 2026-05-23
feat: toBuilder() + DSL writeback for loaded workspaces (#2833)

davydkov:
Closes #2833.

## Summary

Adds programmatic enrichment + DSL writeback for an existing LikeC4 workspace — the literal API from @galuszkak's request, plus an opt-in edit mode for @georgecrawford's "load → patch → write back" use case.

### New public API

- **`LikeC4.parsedModel(project?)`** — exposes the pre-compute parsed model on the `LikeC4` instance.
- **`LikeC4.toBuilder(project?)`** — returns a `Builder` seeded from the parsed workspace; chain `.model(...)` / `.deployment(...)` / `.views(...)` to enrich.
- **`LikeC4.toDSL(project?)`** — renders the parsed model back to formatted `.c4` source.
- **`writeDSL(likec4, dir, opts?)`** — Node-only helper (re-exported from `likec4`) that writes the rendered DSL to disk.
- **`Builder.fromParsed(data)`** — overloaded: typed `Aux` (e.g. `builder.build()` output) preserves types via `Types.FromAux<A>`; data with `UnknownParsed` Aux (the common case from disk) yields a runtime-only Builder.
- **`Builder.fromParsed(...).allowOverwrites()`** — opt-in mode where same-FQN re-declaration *replaces* the existing entry; default mode stays strict and throws on duplicates.
- **`builder.specification(spec)`** — instance method that extends a builder with additional kinds / tags / metadata keys, with `Types.Merge<A, B>` unioning the slots so kinds accumulate across chained `.specification(...)` calls instead of intersecting to `never`.

### Example

```ts
import { LikeC4, writeDSL } from 'likec4'

const likec4 = await LikeC4.fromWorkspace('./')

const enriched = (await likec4.toBuilder())
  .allowOverwrites()
  .specification({ tags: { autogenerated: { color: '#999' } } })
  .model(({ component }, _) =>
    _(component('cloud.new-service', { tags: ['autogenerated'] })),
  )
  .toLikeC4Model()

await writeDSL(likec4, './generated')   // emits ./generated/model.c4
```

### Caveat

The DSL round-trip (`fromWorkspace` → `toDSL` → `fromSource`) is intentionally **lossy**: comments, source positions and original formatting are not preserved. Documented prominently in JSDoc and the user-facing docs.

### Implementation notes

- DSL emission reuses the existing `@likec4/generators/likec4` generator (already in production via `packages/language-server/src/model-change/viewChange.ts`).
- `writeDSL` lives in `@likec4/language-services/node`, so the browser bundle of `language-services` doesn't pull in `node:fs`.
- Integration tests use `memfs` (no real disk), mirroring the language-server `viewChange.spec.ts` pattern.
- `__addElement` is strict by default; the `allowOverwrites()` flag is propagated through `clone()` so chained `.with(...)`, `.model(...)`, etc. preserve the mode.

### Docs

New "Enriching a loaded workspace" section in `apps/docs/.../tooling/model-api.mdx` covering each API, the lossy round-trip caveat, and a type-safety note.

### Test plan

- [x] `pnpm exec tsc --build` clean across core, language-services, likec4, generators
- [x] `Builder.fromParsed.spec.ts` — 9 tests (round-trip, specification preservation, globals, extend-existing, edit-existing via allowOverwrites, cross-refs via helpers(), compute, merge-specification)
- [x] `Builder-style2.test-d.ts` — type-level assertions for chained `.specification(...)`
- [x] `LikeC4.toBuilder.spec.ts` (language-services) — 6 integration tests using memfs (parsedModel / toBuilder / toDSL / round-trip / writeDSL default + custom filename)
- [x] No regression: `packages/core` 992 tests, `packages/language-services` 6 tests, `packages/likec4` 46 tests all green
- [ ] @galuszkak / @georgecrawford sanity check that the API covers the use cases described in #2833


apps/docs/src/content/docs/tooling/model-api.mdx
```diff
@@ -607,3 +607,193 @@ Check unit tests in our repository for examples:
...
+workspace. You can then chain `.model(...)`, `.deployment(...)` and `.views(...)` to add new content,
+and call `.toLikeC4Model()` (or `.build()`) to get the enriched model.
+
+```ts
+import { LikeC4 } from "likec4"
+
+const likec4 = await LikeC4.fromWorkspace('./')
+
+const enriched = (await likec4.toBuilder())
+  .model(({ system, component }, _) =>
```
galuszkak, line 667: Those have to match naming from specification?

apps/docs/src/content/docs/tooling/model-api.mdx
```diff
@@ -607,3 +607,193 @@ Check unit tests in our repository for examples:
...
+import { Builder } from "@likec4/core/builder"
+
+const builder = Builder.fromParsed(parsedData)
+```
+
+### `Builder.allowOverwrites()` — edit seeded elements
+
+By default the Builder is strict: re-declaring an FQN that is already present throws. Opt in with
+`.allowOverwrites()` when you want to *edit* seeded entries (for example, to update a title or merge in
+new children):
```
galuszkak: Is there a way to query / walktrough those existing? What If I want to find element and add just metadata to couple of elements?

apps/docs/src/content/docs/tooling/model-api.mdx
```diff
@@ -607,3 +607,193 @@ Check unit tests in our repository for examples:
...
+const likec4 = await LikeC4.fromWorkspace('./')
+
+const enriched = (await likec4.toBuilder())
+  .model(({ system, component }, _) =>
+    _(
+      system('monitoring').with(
+        component('grafana'),
+        component('prometheus'),
+      ),
+    )
```
galuszkak, line 673: It would be good to show example how to add new element and do relationship with existing element in DSL and new element from Builder added

georgecrawford:
This is working well for me. Thanks @davydkov! 

If I had a wish, it would be to agree with @galuszkak - in most cases, I want to either create a new element, or find an existing element and add or change specific individual properties. I want to update the local model to reflect changes in an upstream system, but I don't want to overwrite the other properties which already exist in the local model.

But given the primitives you've exposed here, I now have what I need to be able to achieve this in other ways, so I'm happy with this PR as it is.

packages/core/CLAUDE.md
galuszkak, line 1: I would keep AGENTS.md @davydkov - users of other agents platforms ;)

apps/docs/src/content/docs/tooling/model-api.mdx
```diff
@@ -607,3 +607,305 @@ Check unit tests in our repository for examples:
...
+// `system` / `component` are statically known — no cast needed.
+builder.model(({ system, component }, _) =>
+  _(system('monitoring').with(component('grafana'))),
+)
+```
+
+The given specification is **validated** against the loaded model (subset semantics): every kind / tag /
+metadata key you declare must exist in the loaded model, otherwise the call throws — so the types can't
+silently drift away from the DSL. The loaded model may contain *extra* kinds you didn't declare; you
+just won't get typed helpers for them. `mode` defaults to `'editable'`, same as `toBuilder`.
```
galuszkak, line 889: I'm little baffled by this - I understand it has some advantages - but redeclaring everything from specification doesn't sit well with me - isn't there anyway to just do some "autodiscover" of those (like process autogenerating file with types based on DSL, rather than declaring them manually)?

galuszkak (approved):

davydkov:
Hey @galuszkak thank you!
I was on a leave due to some personal issues
I want do minor changes (`.allowOverwrites()` by default) and will release this week

PR #3054, opened 2026-06-09
fix: preserve manual layouts in layouted model

### LINKED ISSUE #2553: CLI - 'codegen react' doesn't respect manual layouts
### Relates to

Diagrams (React)

### Version you are using

1.47.0

### Describe the issue

Generated react components are missing `_layout: "manual"` even when a manual layout exists. The same view renders correctly with the manual layout in a generated static site.

Example generated view:
```
  viewAllSystemsSimple: {
    _type: "element",
    tags: null,
    links: null,
    viewOf: "XXX System",
    _stage: "layouted",
    sourcePath: "model.c4",
    description: null,
    title: "XXX System",
    id: "viewXXXSystem",
    autoLayout: { direction: "TB" },  // <-- autoLayout present, but NOT _layout: "manual"
    ...
  }
```

The layoutedModel() method in /packages/language-server/src/LikeC4LanguageServices.ts is calling layoutAllViews() which returns auto-layouted views without applying manual layouts from snapshot files. Meanwhile, the diagrams() method in the same codebase correctly calls withLayoutType(diagram, likeC4Model, 'manual') on each view, which applies manual layouts from .likec4/*.snap files and sets the _layout: 'manual' property.

I'm not positive, but believe the fix could be as simple as changing layoutedModel() to use diagrams() instead of layoutAllViews():

```
  -    const layouted = await this.views.layoutAllViews(projectId, cancelToken)
  +    // Use diagrams() which correctly applies manual layouts via withLayoutType()
  +    const diagrams = await this.views.diagrams(projectId, cancelToken)
       return LikeC4Model.create({
         ...model.$data,
         _stage: 'layouted' as const,
         views: pipe(
  -        layouted,
  -        map(prop('diagram')),
  +        diagrams,
           indexBy(prop('id')),
         ),
       })
```

_No response_

ckeller42:
Fixes #2553

## Summary
- Preserve manual layout positions when `layoutedModel()` is used for generated model data.
- Keep real manual-layout drift reasons, and mark applied no-drift manual views as resolved.
- Add regression coverage for manual-only and mixed manual/auto view models.

## Validation
- Red before fix: manual layout was returned as auto layout in `layoutedModel()`.
- `pnpm --filter @likec4/language-server test -- src/views/LikeC4Views.spec.ts -t "layoutedModel"`
- LikeC4 license header check: passed.
- Cursor: no blocking findings after fixes.
- Claude: requested mixed-view regression coverage; added.
- CodeRabbit: 0 issues.

## Screenshots
Not applicable: this fixes generated layouted model data and has no static visual UI delta.

### DIFF
--- a/.changeset/fix-layouted-model-manual-layouts.md
+++ b/.changeset/fix-layouted-model-manual-layouts.md
@@ -1,0 +1,7 @@
+---
+'@likec4/language-server': patch
+---
+
+Preserve manual layout positions when generating layouted models.
+
+Fixes [#2553](https://github.com/likec4/likec4/issues/2553)
--- a/packages/language-server/src/LikeC4LanguageServices.ts
+++ b/packages/language-server/src/LikeC4LanguageServices.ts
@@ -1,3 +1,10 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2023-2026 Denis Davydkov
+// Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Portions of this file have been modified by NVIDIA CORPORATION & AFFILIATES.
+
 import type { LikeC4ProjectConfig } from '@likec4/config'
 import {
   type LayoutedView,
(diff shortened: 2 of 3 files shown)


ckeller42:
@coderabbitai review

PR #3055, opened 2026-06-09
fix: inherit dynamic relationship kind styles

### LINKED ISSUE #2916: Dynamic view uses relationship styles from specification (still not working)
### Relates to

LikeC4 Language

### Version you are using

1.55.1

### Describe the issue

related to #2797 

Hi @davydkov, @flts

Thanks for the fix — I've tested it against the latest version, but the issue persists (only technology is working; the other relationship properties are still broken). Details see [original issue](#2797)

After further investigation, I narrowed it down to the following:

Relationship defined only in the dynamic view: the visual styles are ignored
Relationship defined in the **model too**: all styles are applied correctly

```
specification {
  element e { style { size xs } } 

  relationship r { line solid; color blue; technology 'T' }
}

model {
  A = e 'A' 
  B = e 'B' 
  C = e 'C' 

  A .r B  'A to B via model'
}

views {
    dynamic view dynamicView {
      include *
      A .r B 'A to B\nreused from model'
      B .r C 'B to C\nvia dynamic view'
    }
}
```

Image

https://playground.likec4.dev/share/Wfi_gfWbuU/

ckeller42:
Fixes #2916

## Summary
- Inherit relationship-kind styles for dynamic view relationships when the model relationship omits explicit style fields.
- Covers color, line, head, tail, and the existing technology inheritance path.
- Add tests for single relationships, shared kind specs, conflicting kind specs, and explicit arrow overrides.

## Validation
- Red before fix: dynamic relationship kind styles missed `head`/`tail` and fell back to default arrows.
- `pnpm --filter @likec4/core test -- src/compute-view/dynamic-view`
- LikeC4 license header check: passed.
- Cursor: no blocking findings after fixes.
- Claude: requested override/conflict coverage; added.
- CodeRabbit: 0 issues.

## Screenshots
Not applicable: this fixes dynamic-view relationship style data and has no static visual UI delta.

### DIFF
--- a/.changeset/fix-dynamic-relationship-kind-styles.md
+++ b/.changeset/fix-dynamic-relationship-kind-styles.md
@@ -1,0 +1,7 @@
+---
+'@likec4/core': patch
+---
+
+Inherit relationship kind styles in dynamic views.
+
+Fixes [#2916](https://github.com/likec4/likec4/issues/2916)
--- a/packages/core/src/compute-view/dynamic-view/__test__/step-fields.spec.ts
+++ b/packages/core/src/compute-view/dynamic-view/__test__/step-fields.spec.ts
@@ -1,3 +1,10 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2023-2026 Denis Davydkov
+// Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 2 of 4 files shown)


ckeller42:
@coderabbitai review

davydkov (approved):

Issue #3076, opened 2026-06-29
Links with label shows both label and URL


AndersHogqvist:
### Relates to

Diagrams (React)

### Version you are using

1.58.0

### Describe the issue

When adding a label to a link (e.g. `link https://likec4.dev 'LikeC4'`) I'd expect that only the label was shown/clickable and I think that was the original implementation as well according to the pull request that implemented this (https://github.com/likec4/likec4/pull/940). However, the result is the title (not clickable) followed by the URL.

### If possible, include link to the playground with example

https://playground.likec4.dev/share/tkU5qNyeYF/

PR #3077, opened 2026-06-29
Fix link with label showing both label and url

AndersHogqvist:
This fixes issue #3076 where a link with a label was showing both the label and the URL. Now, if a label is provided, only the label will be displayed, and the URL will be hidden.

## Checklist

- [x] I've verified `pnpm typecheck` and `pnpm test`. **(found 12 errors but none related to my update)**
- [ ] I've added [changesets](https://changesets-docs.vercel.app/) (you can use `/changeset-generator` SKILL).


davydkov (approved):
Hey @AndersHogqvist thank you for the fix!

PR #3095, opened 2026-07-15
docs: add canonical AGENTS.md instructions

ckeller42:
## Summary

- Add a canonical `AGENTS.md` for shared repository instructions.
- Replace `CLAUDE.md` with the one-line `@AGENTS.md` adapter and remove duplicated nested Claude guidance.
- Add `pnpm check:agent-instructions` plus CI coverage to prevent adapter drift and duplicate instruction sources.
- Keep the changeset-generator agent aligned with the canonical instruction source.

## Validation

- `pnpm check:agent-instructions`
- `dprint check` on touched instruction, workflow, script, and package files
- `git diff --check origin/main...HEAD`
- GitHub CI on latest pushed commit

## Notes

- No changeset: repository guidance and CI-only change; no published package behavior change.

## Checklist

- [x] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x] I've rebased my branch onto `main` **before** creating this PR.
- [x] I've added tests to cover my changes (if applicable).
- [x] I've verified `pnpm typecheck` and `pnpm test`.
- [ ] I've added [changesets](https://changesets-docs.vercel.app/) (not applicable: docs/CI-only repository instruction change).
- [x] My change requires documentation updates.
- [x] I've updated the documentation accordingly (or will do in follow-up PR).

PR #3096, opened 2026-07-15
docs: add LikeC4 developer workflow skills

ckeller42:
Follow-up to #3095.

## Summary

- Add repo-local skills for GitHub PR triage, issue reproduction, CLI codegen regressions, and project config workflows.
- Capture repeated LikeC4 workflow edge cases so agents can use repeatable repo-specific guidance instead of rediscovering the same commands and validation matrix.
- Keep the AGENTS instruction policy in AGENTS.md and the existing validator rather than adding a separate skill for mechanical adapter rules.

## Review notes

- Replaced hard-coded user examples with login placeholders.
- Added pagination/open filtering and exact-location guidance for issue mention triage.
- Added codegen edge cases for stable output naming and synthetic absolute/Windows path coverage.
- Added project-config guidance for workspace startup scanning, where manual document injection can miss dropped include-path files.

## Validation

- quick_validate.py passed for all four new skills
- pnpm exec dprint check on all four new SKILL.md files
- pnpm check:agent-instructions
- git diff --check

### DIFF
--- a/.agents/skills/likec4-cli-codegen-regression/SKILL.md
+++ b/.agents/skills/likec4-cli-codegen-regression/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: likec4-cli-codegen-regression
+description: Use when changing or reviewing LikeC4 CLI code generation, exporters, output file naming, include-path codegen behavior, or `likec4 gen` commands. Trigger for bugs around `--outdir`, generated files escaping output folders, dot/d2/mermaid/plantuml output, or external include-path views.
+---
+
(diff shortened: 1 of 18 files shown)


davydkov (commented):
packages/diagram/src/likec4diagram/xyflow-sequence/CLAUDE.md
I'd prefer to keep for now

ckeller42:
Agreed, kept it.

It now only imports root `AGENTS.md` and points Claude to the xyflow-sequence section, so the detailed rules still have a single source of truth.

davydkov (approved):

PR #3107, opened 2026-07-20
fix(lsp): normalize project exclude path matching

ckeller42:
## Summary

Fixes #2546.

This fixes project exclude matching when LikeC4 receives Windows drive-letter file URIs, for example `file:///C%3A/repo/architecture/test.c4`. Exclude matching now uses decoded URI paths and normalizes only the leading Windows drive letter, so POSIX paths remain case-sensitive.

It also fixes workspace startup so configured `include.paths` are scanned when projects are discovered.

## Examples covered

- `exclude: ['test.c4']` excludes `file:///C%3A/repo/architecture/test.c4`
- `exclude: ['**/wrong.c4']` excludes nested Windows-drive URI paths
- `exclude: ['node_modules']` excludes directory matches under a Windows-drive project root
- excludes also apply to configured include paths
- workspace excludes still take precedence over project ownership/inclusion
- startup scans include-path directories and deduplicates overlapping workspace/include scans

## Validation

- `pnpm --filter @likec4/language-server generate`
- `pnpm --filter @likec4/language-server exec vitest run --no-isolate src/workspace/ProjectsManager.spec.ts src/workspace/WorkspaceManager.spec.ts`
- `pnpm --filter @likec4/language-server typecheck`
- `pnpm exec dprint check packages/language-server/src/workspace/ProjectsManager.ts packages/language-server/src/workspace/ProjectsManager.spec.ts packages/language-server/src/workspace/WorkspaceManager.ts packages/language-server/src/workspace/WorkspaceManager.spec.ts .changeset/twelve-mangos-yell.md`
- `git diff --check --cached`

## Checklist

- [x] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x] I've rebased my branch onto `main` **before** creating this PR.
- [x] I've added tests to cover my changes (if applicable).
- [x] I've verified `pnpm typecheck` and `pnpm test`.
- [x] I've added [changesets](https://changesets-docs.vercel.app/) (you can use `/changeset-generator` SKILL).
- [ ] My change requires documentation updates.
- [ ] I've updated the documentation accordingly (or will do in follow-up PR).

Documentation updates are not applicable for this bug fix; behavior is covered by regression tests and a patch changeset.

Issue #2546, opened 2026-01-12
Exclusions in project definition stopped working in 1.47.0


kozziou:
### Relates to

LikeC4 Language

### Version you are using

1.47.0

### Describe the issue

Files that were defined as exclusions in the project definition file (likec4.config.json) stopped being excluded in 1.47.0.

davydkov:
@kozziou Ouch, might be regression.

Can you share details - platform and your exclusions?

kozziou:
The configuration file in question looks like this:

```
{
    "name": "architecture-repository",
    "title": "Repozytorium architektury",
    "exclude": [
        "**/proj-template.c4**"
    ]
}
```

It used to exclude this file: /architecture/projects/proj-template.c4 (it defines a template for a project view). Now I see the template view listed in the navigation menu. Also, as a test, I've added model elements to this file and they started appearing on other views.

I noticed the behavior in VS Code extension (Windows) but I confirmed it also happens in local server and html export.

ckeller42:
This was fixed in commit 39dcb7732 (Dec 2025) — `fix: exclude option in project configuration`. The `updateIncludesExcludes` method in `ProjectsManager.ts` now correctly applies picomatch patterns from `likec4.config.json` exclude arrays. Tests exist in `ProjectsManager.spec.ts` (line 344+). Please update to the latest version and confirm.

kozziou:
As of 1.55.1 the issue still persists for me.

My likec4.config.json is the same as in my previous comment. The template view defined in this file is still visible in the navigation menu (both in VS Code Preview pane and in the exported html).

kozziou:
Can you please take another look at this issue? From my perspective this functionality is not working at all and is seriously impacting my multi-project workflow.

kozziou:
@davydkov, @ckeller42

I can't, for the life of me, get it to work, even on 1.59.0.

A simple test:
- likec4.config,json
```
{
    "$schema": "https://likec4.dev/schemas/config.json",
    "name": "architecture-repository",
    "title": "Repozytorium architektury",
    "exclude": [
        "test.c4"
    ],
    "implicitViews": false,
    "landingPage": {
        "redirect": true
    }
}
```
- test.c4 (in the same directory)
```
views {
    view test {
        title '!TEST!'
        
        include *
    }
}
```
- In the preview in VS and in the live server (likec4 start) the view with the title "!TEST!" is still visible in the navigation and can be opened.
- I've also tried 
```
"**test.c4**"
```
[...]

PR #3108, opened 2026-07-20
test(lsp): prove project excludes on Windows paths

ckeller42:
## Summary

Follow-up to #3107.

References #2546.

Adds a Windows-only regression test for the reporter's simple project config case:

- `exclude: ["test.c4"]`
- `test.c4` in the same folder as the project config
- excluded file defines a visible view titled `!TEST!`

The test uses real Windows path input on `windows-latest` and verifies that the excluded file is not included in project documents and does not contribute the `!TEST!` view to the computed model.

## CI evidence

- Red repro on `main`: [`checks / ⊞ windows build` failed on temporary PR #3109](https://github.com/likec4/likec4/actions/runs/29759878013/job/88411537880).
- Green fixed run before the guard-only review cleanup: [`checks / ⊞ windows build` passed on PR #3108](https://github.com/likec4/likec4/actions/runs/29759825678/job/88411361643).

## Validation

- `pnpm --filter @likec4/language-server exec vitest run --no-isolate src/workspace/ProjectsManager.spec.ts -t "same-directory files"`
- `pnpm --filter @likec4/language-server exec vitest run --no-isolate src/workspace/ProjectsManager.spec.ts`
- `pnpm exec vitest run packages/language-server/src/workspace/ProjectsManager.spec.ts`
- `pnpm --filter @likec4/language-server typecheck`
- `pnpm exec dprint check packages/language-server/src/workspace/ProjectsManager.spec.ts`
- `git diff --check`

## Checklist

- [x] I've thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [ ] I've rebased my branch onto `main` **before** creating this PR.
- [x] I've added tests to cover my changes (if applicable).
- [x] I've verified `pnpm typecheck` and `pnpm test`.
- [ ] I've added [changesets](https://changesets-docs.vercel.app/) (you can use `/changeset-generator` SKILL).
- [ ] My change requires documentation updates.
- [ ] I've updated the documentation accordingly (or will do in follow-up PR).

Rebase onto `main` is not applicable because this is intentionally stacked on #3107. A changeset and documentation updates are not applicable because this PR is test-only.

PR #3109, opened 2026-07-20
test(lsp): reproduce project exclude failure on Windows

ckeller42:
## Summary

Temporary CI repro for #2546.

This branch intentionally adds only the Windows regression test from the stacked follow-up PR, but does not include the fix from #3107.

Expected result: `checks / ⊞ windows build` fails because `exclude: ["test.c4"]` does not exclude the same-directory file on Windows paths before #3107.

This PR is only for red CI evidence and should be closed after the evidence is linked from the stacked follow-up PR.


ckeller42:
Closing this temporary repro PR. The red Windows CI evidence is preserved and linked from #3108.

PR #3146, opened 2026-07-25
fix(language-server): do not duplicate keyword in completions

### LINKED ISSUE #3091: Language server reports model as modelmodel
### Relates to

CLI

### Version you are using

1.58.0

### Describe the issue

The language server presents a `model` entity as `modelmodel`.

Image

As seen on Zed with the LikeC4 extension 0.0.1.

_No response_

farhan523:
Closes #3091

## Problem

Keyword completions build an `insertText` that **starts with the keyword itself**, but do not set a `textEdit`:

https://github.com/likec4/likec4/blob/v1.59.2/packages/language-server/src/lsp/CompletionProvider.ts#L76-L85

Langium's `fillCompletionItem` then emits an item carrying **both** the `insertText` and a derived `textEdit` whose range replaces the typed token.

Per LSP, a client must prefer `textEdit` — VS Code does, so completions look correct there. Clients that apply `insertText` at the cursor instead insert the keyword a second time. Typing `model` and accepting the suggestion produces `modelmodel`, as reported on Zed.

This affected every snippet/property keyword, not just `model`: `views`, `specification`, `deployment`, `style`, `group`, `title`, `description`, and so on.

## Fix

Provide an explicit `textEdit` for these items so the typed prefix is always replaced, regardless of how the client resolves the item.

The helper delegates to the inherited `buildCompletionTextEdit`, which means the existing fuzzy-match gating is preserved — items are still only offered when the typed prefix matches the keyword. This was important: an earlier attempt that built the range by hand and set `filterText` regressed 10 tests by bypassing that gate.

No behaviour change in VS Code, which already used the `textEdit`.

## Tests

Two tests in `CompletionProvider.spec.ts` assert the emitted payload directly, which the existing tests did not cover (they only ever assert `label`):

- typing `mod` → the `model` item's `textEdit` replaces `[0,0]-[0,3]` with `model {\n\t$0\n}`, and `insertText` is absent
- typing `titl` inside an element → the `title` item's `textEdit` carries `title '$0'`

Both fail before the fix (`expected 'model {\n\t$0\n}' to be undefined`) and pass after.

## Verification

- `pnpm vitest run --project language-server packages/language-server/src/lsp/CompletionProvider.spec.ts` → 19 passed, 1 todo
- Full `--project language-server` suite compared against `main`: same results, no regressions (+2 from the new tests)
- `tsc --noEmit` clean, oxlint clean, dprint clean

### DIFF
--- a/.changeset/fix-completion-duplicate-keyword.md
+++ b/.changeset/fix-completion-duplicate-keyword.md
@@ -1,0 +1,5 @@
+---
+'@likec4/language-server': patch
+---
+
+Fix keyword completions inserting the keyword twice (e.g. `modelmodel`) in editors other than VS Code. Snippet and property completions now provide an explicit text edit that replaces the typed prefix, instead of relying on the client to prefer it over the plain insert text.
--- a/packages/language-server/src/lsp/CompletionProvider.spec.ts
+++ b/packages/language-server/src/lsp/CompletionProvider.spec.ts
@@ -1167,4 +1167,54 @@ describe('LikeC4CompletionProvider', () => {
       ],
     })
   })
+
+  // Issue #3091 - completion items must carry an explicit textEdit,
+  // otherwise clients relying on insertText duplicate the typed keyword ("modelmodel")
+  it('should not duplicate typed keyword in snippet completions', async ({ expect, completion }) => {
+    const text = `mod<|>`
+    await completion({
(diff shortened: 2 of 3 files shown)


davydkov (approved):
Thanks for contribution!

PR #3151, opened 2026-07-26
fix(language-server): wrap document symbol detail in parens to avoid merging with name

### LINKED ISSUE #3091: Language server reports model as modelmodel

bounce12340:
## Summary

Fixes #3091 — in some LSP clients (e.g. Zed), a document symbol's `name` and `detail` are rendered adjacently without a separator. Since `detail` was set to the bare element kind, an element whose name equals its kind (like `model model`) displayed as `modelmodel`.

This wraps the kind in parentheses (`(model)`), so the two stay visually distinct in any client, matching the common convention used by other language servers.

Applied to both logical elements and deployment nodes in `DocumentSymbolProvider`.

## Verification

Added a regression test reproducing the issue with `element model` / `model model`:

```
pnpm vitest run DocumentSymbolProvider
# Test Files  1 passed (1)
#      Tests  2 passed (2)
```

`pnpm typecheck` passes on this branch.

A full `pnpm test` run does not complete in my environment: 20 test files fail to import `@likec4/styles/css` and `prompt-system.generated`, i.e. build artifacts that a plain checkout does not have. Those are missing-package errors, not assertion failures — the run still reports 2528 passing tests and no failing assertions. Flagging it rather than ticking the box below, since I could not verify the whole suite myself.

## Checklist

- [x] I've rebased my branch onto `main` **before** creating this PR. <sup>(current at creation; 3 commits have landed on `main` since — happy to rebase if you'd like it refreshed before merge)</sup>
- [ ] I've verified `pnpm typecheck` and `pnpm test`. <sup>(`typecheck` verified; full `pnpm test` not verifiable locally — see Verification above)</sup>

### DIFF
--- a/.changeset/fix-document-symbol-detail.md
+++ b/.changeset/fix-document-symbol-detail.md
@@ -1,0 +1,5 @@
+---
+"@likec4/language-server": patch
+---
+
+Fix outline entries merging an element's kind into its name, such as `modelmodel`. The kind now renders in parentheses. Fixes [#3091](https://github.com/likec4/likec4/issues/3091)
--- a/packages/language-server/src/lsp/DocumentSymbolProvider.spec.ts
+++ b/packages/language-server/src/lsp/DocumentSymbolProvider.spec.ts
@@ -93,4 +93,32 @@ describe('LikeC4DocumentSymbolProvider', () => {
       },
     ])
   })
+
+  it('should keep element detail from visually merging with name when kind equals name (#3091)', async ({ expect, t, validate }) => {
+    const { document, diagnostics } = await validate(`
+        specification {
+          element model
+        }
+        model {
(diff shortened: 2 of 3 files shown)


davydkov (approved):
Thank you for contribution

davydkov:
@bounce12340 Can you please add a changeset file? (use changeset-generator skill)

bounce12340:
Added the changeset (`.changeset/fix-document-symbol-detail.md`, patch for `@likec4/language-server`) — thanks for the review!

PR #3170, opened 2026-08-02
Add MiniMax auto-detection to the Vite plugin

octo-patch:
Reason: Add MiniMax auto-detection through the OpenAI-compatible adapter path.

Implemented MiniMax detection in `packages/vite-plugin/src/ai/detect-ai.ts` with the `MiniMax-M3` default model, `MiniMax-M2.7` override, and global/CN endpoint selection.
Added focused Vitest coverage for the default endpoint, alternate model, and CN endpoint selection.

Checks:
- pnpm vitest run packages/vite-plugin/src/ai/detect-ai.spec.ts --no-isolate
- pnpm exec tsc -p packages/vite-plugin/tsconfig.json --noEmit --pretty false --skipLibCheck


davydkov (approved):
Thanks for contribution!

PR #3174, opened 2026-08-05
Adding 2 MCP Apps (render-view, preview-view) to the mcp server to show views in AI Chats

kgeilmann:
Hello

As already shown on the discord I extended the mcp server with two mcp app tools:

- render-view: renders an existing view, directly in the AI chat
- preview-view: takes a view definition (in likec4 dsl) and renders exactly that view in the AI chat, this allows to create kind of temporary views on the fly for an existing model

Both tools use the react component, so the views are interactive to some extend.

I tried it with github copilot in vscode (should also work in claude desktop or any other mcp app enabled chat)

Up to now, I'm happy with the functionality, but not so with the tool names, especially with "preview-view". Might be better to call it create-temporary-view, any thoughts about that?

## Checklist


Abhirup0 (commented):
Hey @kgeilmann! Really cool PR, having interactive MCP apps to preview and render diagrams directly inside chat tools is super handy.

Took a look through the code and just wanted to share a couple of small things I noticed while checking `preview-view.ts`, mainly around a few syntax edge cases:

1. **Scoped views (`view of ...`)**:
   In `extractViewId`, the regex is currently `^\s*view\s+([A-Za-z_][\w-]*)`. If someone passes a scoped element view like `view of myElement { ... }` (which is valid LikeC4 syntax without an explicit view ID), it might capture `"of"` as the view ID and then fail on `findView("of")`. We might want to allow an optional `of` in the regex or handle scoped declarations so those don't trip it up.

2. **Input already wrapped in a `views` block**:
   In `wrapAsViewsBlock`, it wraps the incoming DSL in `views { ... }`. If an LLM or user already passes a snippet with the outer block (e.g. `views { view v1 { ... } }`), it could end up generating nested `views { views { ... } }` and fail parsing. A quick check like `trimmed.startsWith('views') ? trimmed : ...` could make it a bit more forgiving for callers.

3. **Version bump in `meta.ts`**:
   Looks like `packages/vscode/src/meta.ts` got bumped to `1.59.2`. Just mentioning in case it was accidental, since changesets usually take care of versioning during releases.

Really clean setup overall with the bundling and model isolation via `fromSources`! Hope this helps.

davydkov (approved):
Hey @kgeilmann 

Very nice contribution!!!

I'm sorry for delay with review, vacation period

PR #3182, opened 2026-08-17
fix(cli): count files matched by validate --file, not only files with errors

### LINKED ISSUE #3180: `validate --file` reports `filteredFiles: 0` for a valid matching file
### Relates to

CLI

### Version you are using

1.59.2

### Describe the issue

`likec4 validate --json --file` reports `filteredFiles: 0` when the selected file exists, belongs to the project, and has no diagnostics.

This makes `filteredErrors: 0` ambiguous: callers cannot distinguish a clean matched file from an unmatched filter.

### Reproduction
Create a `likec4.config.json` file with:
```json
{
  "$schema": "https://likec4.dev/schemas/config.json",
  "name": "filtered-files-repro"
}
```

And then a `model.c4` with:

```
specification {
  element person
}

model {
  person user "User"
}
```

Then run:

```bash
npx --yes likec4@1.59.2 validate \
  --json \
  --no-layout \
  --file model.c4 \
  .
```

### Actual result

```json
{
  "valid": true,
  "errors": [],
  "stats": {
    "totalFiles": 1,
    "totalErrors": 0,
    "filteredFiles": 0,
    "filteredErrors": 0
  }
}
```

### Expected result

`filteredFiles` should be `1` because `model.c4` matched the filter:

```json
{
  "valid": true,
  "errors": [],
  "stats": {
    "totalFiles": 1,
    "totalErrors": 0,
    "filteredFiles": 1,
    "filteredErrors": 0
  }
}
```

The same result occurs when `--file` receives the exact absolute path.

As a control, changing `person user "User"` to the invalid `unknown user "User"` makes the same filter report:

```json
{
  "totalFiles": 1,
  "totalErrors": 1,
  "filteredFiles": 1,
  "filteredErrors": 1
}
```

This suggests `filteredFiles` is derived from files represented in filtered diagnostics rather than source files matched by `--file`.

It would be useful for filter matching to be calculated independently of diagnostics, with `filteredFiles` reporting matched source files and `filteredErrors` reporting diagnostics from those files.

_No response_

luantaraschi:
## Checklist

Fixes #3180.

`validate --file` built `filteredFiles` out of the diagnostics that survived the filter:

```ts
const filteredErrors = fileFilter ? allErrors.filter(e => fileFilter.some(...)) : allErrors
const filteredFileSet = new Set(filteredErrors.map(e => e.file))
...
filteredFiles: fileFilter ? filteredFileSet.size : totalFiles,
```

A file that matched `--file` but carried no diagnostics contributed nothing to that set, so it reported `0`. The match predicate was only ever applied to errors, never to the documents.

No documentation change is needed, because the docs already describe the behaviour this PR implements. `skills/likec4-dsl/SKILL.md` says `"filteredFiles": 1, // Number of files that match the --file filter` and, a few lines below, "Selfcheck that `filteredFiles` matches the number of files you passed to `--file`". This brings the code to the documented contract rather than changing it.

### What changed

`LikeC4.documentPaths()` in `packages/language-services/src/common/LikeC4.ts`, right next to `documentCount()` that this same stats block already calls. There was no public way to reach the document paths, only their count.

The stats computation moved out of the yargs handler into `packages/likec4/src/cli/validate/result.ts` as a pure `buildValidateResult`, following the shape of `cli/codegen/handler.ts`, so it can be tested without spinning up a workspace. `filteredFiles` is now counted over the documents.

I left `documentCount()` alone instead of folding it into the new method, to keep the diff narrow. Happy to collapse the two if you prefer.

### Against the reproduction in the issue

Same workspace and same command as the issue, run from the workspace folder:

```
likec4 validate --json --no-layout --file model.c4 .
```

Before, `"filteredFiles": 0`. After, `"filteredFiles": 1`, with the rest of the payload unchanged.

I also checked the edge case the skill documents. Passing `--file model.c4 --file views.c4 --file likec4.config.json` still reports `filteredFiles: 2`, because the config file is not a parsed document, so that note stays accurate.

### Testing

`packages/likec4/src/cli/validate/result.spec.ts`, five cases: a matched clean file, a mixed set where only some files carry errors, a file left out by the filter, a filter given as a relative path, and the no-filter path. Four of the five fail on `main` and all five pass with the change. The no-filter case passes on both, which is the point: that path was already right.

Ran in a Linux container on Node 22.23.1 with pnpm 11.15.0, so the `.tool-versions` toolchain: `pnpm install`, `pnpm generate`, `pnpm exec tsc --build`, then `pnpm typecheck` (26 tasks, all successful) and `pnpm test` (274 files, 2767 passed, 1 expected fail, 23 skipped, 4 todo). `dprint check` and `oxlint` are clean on the four touched files.

I did not run `pnpm test:e2e`.

Disclosure: I used an AI coding assistant while working on this. Every command and result reported above I ran and checked myself.

### DIFF
--- a/.changeset/fix-validate-filtered-files-count.md
+++ b/.changeset/fix-validate-filtered-files-count.md
@@ -1,0 +1,6 @@
+---
+'likec4': patch
+'@likec4/language-services': patch
+---
+
+`likec4 validate --file` now counts every file matched by the filter in `filteredFiles`, including files with no diagnostics. It previously counted only files that carried errors, so a clean matched file reported `filteredFiles: 0`.
--- a/packages/language-services/src/common/LikeC4.ts
+++ b/packages/language-services/src/common/LikeC4.ts
@@ -332,6 +332,13 @@ Please specify a project folder`)
   }
 
   /**
+   * Returns the file system paths of the parsed documents in the workspace
+   */
+  documentPaths(): string[] {
+    return [...this.LangiumDocuments.userDocuments].map(doc => doc.uri.fsPath)
+  }
+
+  /**
    * Subscribe to model updates
    * @param listener Function called when the model is updated - receives the project ID that was updated
    * @returns A function to dispose the listener
--- a/packages/likec4/src/cli/validate/index.ts
+++ b/packages/likec4/src/cli/validate/index.ts
@@ -4,27 +4,8 @@ import k from 'tinyrainbow'
 import type * as yargs from 'yargs'
 import { createLikeC4Logger, startTimer } from '../../logger'
 import { path, project } from '../options'
-
-interface DiagnosticItem {
-  message: string
-  file: string
-  line: number
-  range: {
-    start: { line: number; character: number }
-    end: { line: number; character: number }
-  } | null
-}
-
-interface ValidateResult {
-  valid: boolean
-  errors: DiagnosticItem[]
-  stats: {
-    totalFiles: number
-    totalErrors: number
-    filteredFiles: number
-    filteredErrors: number
-  }
-}
+import type { DiagnosticItem } from './result'
+import { buildValidateResult } from './result'
 
 const validateCmd = (yargs: yargs.Argv) => {
   return yargs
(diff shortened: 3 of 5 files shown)


davydkov (approved):
Thanks for contribution

PR #3187, opened 2026-08-18
fix: preserve automatic edge routing when moving a node

### LINKED ISSUE #3184: Moving a node can turn generated edge curves into manual waypoints
### Relates to

Diagrams (React)

### Version

1.59.2

### What happened

Moving a node can quietly turn LikeC4's generated edge curves into authored waypoints. After saving, edges that had never been manually routed now have persisted `controlPoints`.

This became very visible in a larger diagram: I moved one node, saved, and later found tall arches and seemingly random bends on a number of edges. The snapshot had frozen points from the old generated Bézier curves as if I had placed them myself.

### How to reproduce

1. Open a view using generated layout, or reset a manually laid-out view to generated layout.
2. Enter Edit mode.
3. Move a node with one or more automatically routed incident edges.
4. Save and inspect the resulting layout snapshot.

Edges without authored waypoints can acquire `controlPoints` derived from their rendered curve.

### What I expected

Moving a node should preserve the distinction between automatic and manually routed edges:

- If an edge had no authored control points, it should remain automatic after the move.
- If the user had explicitly added waypoints, those points should move with the node as they do today.

I traced the behavior to the edge modifier falling back to generated Bézier control points and then writing that fallback back as authored `controlPoints`. Keeping track of whether the points were originally explicit avoids promoting the generated curve.

Phrogz:
Fixes #3184.

## What changed

- Keep automatically routed relationships represented by generated path points when a connected node moves.
- Continue moving explicitly authored control points relative to the dragged endpoint.
- Stop snapshot serialization from promoting generated Bézier handles into authored waypoints.
- Keep generated edge coordinates finite when connected nodes have the same center.

This prevents ordinary node moves from silently freezing generated curves into manual routing that can later produce large, unexpected bends.

## Validation

- `pnpm generate`
- `pnpm typecheck`
- `ulimit -n 65536 && CI=true pnpm test` (2,766 passed, 1 expected failure, 23 skipped, 4 todo)
- Focused diagram regression tests
- Type-aware oxlint on all changed TypeScript files

## Checklist

- [x] I have thoroughly read the latest [contribution guidelines](https://github.com/likec4/likec4/blob/main/CONTRIBUTING.md).
- [x] I have rebased my branch onto `main` **before** creating this PR.
- [x] I have added tests to cover my changes.
- [x] I have verified `pnpm typecheck` and `pnpm test`.
- [x] I have added a [changeset](https://changesets-docs.vercel.app/).
- [ ] I have updated the documentation accordingly (or will do in a follow-up PR).

### DIFF
--- a/packages/diagram/src/likec4diagram/state/createViewChange.ts
+++ b/packages/diagram/src/likec4diagram/state/createViewChange.ts
@@ -7,7 +7,6 @@ import type {
 import { getNodeDimensions } from '@xyflow/system'
 import { hasAtLeast, map } from 'remeda'
 import { calcViewBounds } from '../../utils/view-bounds'
-import { bezierControlPoints, isSamePoint } from '../../utils/xyflow'
 import type { DiagramContext } from './types'
 
 export function createViewChange(
@@ -24,7 +23,6 @@ export function createViewChange(
   } = parentContext
 
   const { nodeLookup, edgeLookup } = xystore.getState()
-  const movedNodes = new Set<string>()
 
   const nodes = map(view.nodes, (node): DiagramNode => {
     const internal = nodeLookup.get(node.id)
@@ -35,14 +33,6 @@ export function createViewChange(
     const xynodedata = xynodes.find(n => n.data.id === node.id)?.data ?? internal.data
     const position = internal.internals.positionAbsolute
     const { width, height } = getNodeDimensions(internal)
-
-    const isChanged = !isSamePoint(position, node)
-      || node.width !== width
-      || node.height !== height
-
-    if (isChanged) {
-      movedNodes.add(node.id)
-    }
 
     return {
       ...node,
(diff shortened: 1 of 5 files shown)


packages/diagram/src/likec4diagram/state/createViewChange.ts
```diff
@@ -65,12 +55,7 @@ export function createViewChange(
       return edge
     }
     const data = xyedge.data
-    let controlPoints = data.controlPoints ?? []
-    const sourceOrTargetMoved = movedNodes.has(xyedge.source) || movedNodes.has(xyedge.target)
-    // If edge control points are not set, but the source or target node was moved
-    if (controlPoints.length === 0 && sourceOrTargetMoved) {
-      controlPoints = bezierControlPoints(data.points)
-    }
```
davydkov, line 73: Be honest, I do not recall what case I tried to solve here 😄

davydkov (approved):

davydkov:
Hey @Phrogz 

I'm about to revert the changes as now we get this issue

https://github.com/user-attachments/assets/67a22afb-ef48-4447-9a8b-3ca9744fe050

PR #3200, opened 2026-08-31
fix(diagram): walkthrough crash when clicking a step in diagram variant

### LINKED ISSUE #3201: Walkthrough crashes when clicking another step in a diagram-variant dynamic view
### Relates to

Diagrams (React)

### Version you are using

1.59.2

### Describe the issue

In a `diagram`-variant dynamic view, clicking another step while a walkthrough is active crashes the diagram:

```
Error: Expected seq-step edge, but got "relationship"
```

Steps to reproduce, in the read-only viewer (`likec4 start`, or a site built with `likec4 build`):

1. Open any dynamic view — the default `diagram` variant
2. Click a step arrow once, it gets selected
3. Click the same arrow again, the walkthrough starts
4. Click a different step arrow — the diagram crashes with the error above

Keyboard navigation (`←` / `→`) and `Esc` are unaffected, so the walkthrough itself works. Only switching steps by clicking fails.

The walkthrough state handles `xyflow.edgeClick` in two branches. The second one, taken for any step other than the active one, asserts the ReactFlow edge type — `packages/diagram/src/likec4diagram/state/machine.state.ready.walkthrough.ts`:

```ts
invariant(event.edge.type === 'seq-step', `Expected seq-step edge, but got "${event.edge.type}"`)
```

Steps are rendered as `seq-step` edges only in the `sequence` variant. In the `diagram` variant they are `relationship` edges, so the assertion never holds there.

The assertion arrived in 86282c12 and has been released since v1.59.0 (v1.58.0 does not contain it). Before that, the same branch checked the step id instead of the edge type and worked in both variants.

I could not reproduce it in the playground: with the editor open the diagram is not read-only, and the `xyflow.edgeClick` → `walkthrough.start` transition is guarded by `enabled: Readonly` (`machine.state.ready.idle.ts`). The read-only viewer reproduces it with the steps above.

wataryooou:
Fixes #3201

## Problem

In a `diagram`-variant dynamic view, clicking another step while a walkthrough is active crashes the diagram:

```
Error: Expected seq-step edge, but got "relationship"
```

Steps to reproduce in the read-only viewer (`likec4 start`), on 1.59.2:

1. Open any dynamic view — the default `diagram` variant
2. Click a step arrow once, it gets selected
3. Click the same arrow again, the walkthrough starts
4. Click a different step arrow — the diagram crashes with the error above

Keyboard navigation (`←` / `→`) and `Esc` are unaffected, so the walkthrough itself works. Only switching steps by clicking fails.

## Cause

The walkthrough state handles `xyflow.edgeClick` in two branches. The second one, taken for any step other than the active one, asserts the ReactFlow edge type:

```ts
invariant(event.edge.type === 'seq-step', `Expected seq-step edge, but got "${event.edge.type}"`)
```

Steps are rendered as `seq-step` edges only in the `sequence` variant. In the `diagram` variant they are `relationship` edges, so the assertion never holds there.

The assertion arrived in 86282c12 and has been released since v1.59.0 (v1.58.0 does not contain it). Before that, the same branch checked the step id instead of the edge type and worked in both variants. The rest of the handler only reads `event.edge.data.id`, so it is variant-agnostic.

## Fix

Assert on the step id, the way `walkthrough.step` already does in the same file:

```ts
const stepId = event.edge.data.id
invariant(isStepPath(stepId), `Expected step edge, but got "${stepId}"`)
```

## Tests

`machine.state.ready.walkthrough.spec.ts` drives the machine into a walkthrough on a `diagram`-variant dynamic view and asserts that

- clicking another step moves the walkthrough to it (this threw before the fix)
- clicking the active step keeps the walkthrough where it is

## Checklist

### DIFF
--- a/.changeset/fix-walkthrough-edge-click-in-diagram-variant.md
+++ b/.changeset/fix-walkthrough-edge-click-in-diagram-variant.md
@@ -1,0 +1,5 @@
+---
+'@likec4/diagram': patch
+---
+
+Fix crash when clicking another step during a walkthrough in a `diagram`-variant dynamic view. The click handler asserted the ReactFlow edge type, which is `seq-step` only in the `sequence` variant, and now checks the step id instead. Fixes [#3201](https://github.com/likec4/likec4/issues/3201).
--- a/packages/diagram/src/likec4diagram/state/machine.state.ready.walkthrough.spec.ts
+++ b/packages/diagram/src/likec4diagram/state/machine.state.ready.walkthrough.spec.ts
@@ -1,0 +1,172 @@
+import type { LayoutedDynamicView, NonEmptyArray, Point } from '@likec4/core/types'
+import { scalar } from '@likec4/core/types'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { createActor, fromCallback } from 'xstate'
+import { DefaultFeatures } from '../../context/DiagramFeatures'
+import type { XYFlowInstance, XYStoreApi } from '../../hooks/useXYFlow'
+import type { Types } from '../types'
+import { diagramMachine } from './machine'
+
+// One step of the view, as the layouter leaves it.
+const mockViewEdge = (stepnum: number) => ({
+  id: scalar.EdgeId(`step-0${stepnum}`),
+  parent: null,
+  source: scalar.NodeId('source-node'),
+  target: scalar.NodeId('target-node'),
+  label: `step ${stepnum}`,
+  relations: [],
+  color: 'primary' as const,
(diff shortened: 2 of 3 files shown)


wataryooou:
Skipping this one: the repo's dprint config sorts imports by module path, and `dprint check` fails on the suggested order (it moves the two type-only imports back).
`importDeclaration.sortTypeOnlyImports: "first"` applies within an import declaration, not across declarations. 
The neighbouring `machine.state.navigating.spec.ts` uses the same interleaved order.

davydkov (approved):

PR #3204, opened 2026-08-31
fix(cli): honour LIKEC4_WORKSPACE in `likec4 mcp`

### LINKED ISSUE #3203: likec4 mcp ignores LIKEC4_WORKSPACE and silently serves an empty model
### Relates to

CLI

### Version you are using

1.59.2

### Describe the issue

`likec4 mcp` documents `LIKEC4_WORKSPACE` in its own help text but never reads it. The value is discarded and the workspace always falls back to the current directory.

```
$ likec4 mcp --help
  path  <directory> with LikeC4 sources (default is current directory or 'LIKEC4_WORKSPACE' env)
```

The failure is silent. When the current directory holds no `.c4` files the server still starts, `initialize` succeeds and every tool answers — against an empty model. The only signal is a `WARN likec4.server.workspace No likec4 documents found in workspace` on stderr, which an MCP client does not surface. From the client's side LikeC4 simply appears to know nothing about the architecture.

This matters for MCP specifically because the server is spawned by a client, not typed into a shell, so the working directory is whatever the client chose and passing the workspace through the environment is the natural thing to do — it is what the LikeC4 docs recommend for editors.

### Reproduction

With any workspace that contains `.c4` files, from a directory that does not:

```console
$ cd /tmp   # no .c4 files here
$ LIKEC4_WORKSPACE=/path/to/workspace likec4 mcp
# -> list-projects returns 0 sources

$ likec4 mcp /path/to/workspace
# -> list-projects returns 40 sources

$ LIKEC4_WORKSPACE=/path/to/workspace likec4 validate
# -> ✓ Valid (40 files)
```

Same binary, same path, same directory. `validate` honours the variable, `mcp` does not. `build`, `codegen`, `format`, `serve` and `sync` all behave like `validate`.

The variable appears to work when it is set to `.`, but that is a coincidence: `.` happens to equal the hardcoded fallback.

### Cause

`packages/likec4/src/cli/options.ts` puts the environment variable in the default of the shared `path` positional:

```ts
export const path = {
  type: 'string',
  desc: '<directory> with LikeC4 sources (default is current directory or \'LIKEC4_WORKSPACE\' env)',
  normalize: true,
  default: env['LIKEC4_WORKSPACE'] || '.',
  coerce: resolve,
} as const satisfies PositionalOptions
```

`packages/likec4/src/cli/mcp/index.ts` then overrides that default one line later:

```ts
.positional('path', path)
.default('path', '.', 'current directory')   // <- discards the env-derived default
```

Reduced to yargs alone, with `LIKEC4_WORKSPACE=/from/env`:

```
current code (.default override present):
  path = /current/working/directory
with the override removed:
  path = /from/env
```

### Suggested fix

Drop the `.default('path', …)` call in `packages/likec4/src/cli/mcp/index.ts`. The shared `path` positional already carries both the default and the help text describing it, so removing the override restores the documented behaviour and makes `mcp` consistent with the other commands. Happy to open a PR.

### Side note

`packages/likec4/src/cli/preview/index.ts` never uses the shared `path` positional — it declares its own with `default: resolve('.')`, so `preview` does not support `LIKEC4_WORKSPACE` either. Its help text does not advertise the variable, so this is an inconsistency rather than the same bug, and I left it out of the suggested fix. Let me know if you would like `preview` aligned as well.

tnucera:
Fixes #3203.

`likec4 mcp` advertises `LIKEC4_WORKSPACE` in its own help text but never reads it, so the workspace always falls back to the current directory.

The shared `path` positional in `packages/likec4/src/cli/options.ts` already carries `env['LIKEC4_WORKSPACE'] || '.'` as its default, and `mcp` overrode it one line after applying it:

```ts
.positional('path', path)
.default('path', '.', 'current directory')   // discards the env-derived default
```

Removing the override restores the documented behaviour and makes `mcp` consistent with `build`, `codegen`, `format`, `serve`, `sync` and `validate`, which all honour the variable.

### Why this is worth a patch

The failure is silent. With no sources in the current directory the server still starts, `initialize` succeeds and every tool answers — against an empty model. The only signal is a `WARN likec4.server.workspace No likec4 documents found in workspace` on stderr, which MCP clients do not surface: from the client's side LikeC4 simply appears to know nothing about the architecture.

It hits `mcp` harder than the other commands because the server is spawned by a client rather than typed into a shell, so the working directory is not the user's choice and passing the workspace through the environment is the natural thing to do.

Setting the variable to `.` appears to work, but only because `.` coincides with the hardcoded fallback.

### Verification

Same build, same command, same environment, from a directory holding no `.c4` files:

| | sources resolved |
|---|---|
| before, `LIKEC4_WORKSPACE=<workspace>` | **0** |
| after, `LIKEC4_WORKSPACE=<workspace>` | **40** |
| after, positional `<workspace>` | **40** (unchanged) |
| after, neither set | falls back to cwd (unchanged) |

Measured by driving the stdio server directly — `initialize`, then a `list-projects` tool call — and counting the returned sources.

- `pnpm --filter likec4 typecheck` — clean
- `vitest run --no-isolate` on `packages/likec4` — 11 files, 57 passed, 4 skipped

### Note on `preview`

`packages/likec4/src/cli/preview/index.ts` does not use the shared `path` positional — it declares its own with `default: resolve('.')`, so it does not support `LIKEC4_WORKSPACE` either. Its help text does not advertise the variable, so I treated it as an inconsistency rather than the same bug and left it out to keep this PR minimal. Happy to align it here or in a follow-up if you prefer.

## Checklist

- [x] I've thoroughly read the latest contribution guidelines.
- [x] I've added changesets.

On tests: there is currently no harness covering CLI option parsing — `packages/likec4/src/cli/**` has `serve.spec.ts` and two codegen specs, none of which exercise yargs wiring. I did not want to introduce one unprompted for a one-line fix, but I'm glad to add it if you'd like this covered.

### DIFF
--- a/.changeset/mcp-honour-likec4-workspace-env.md
+++ b/.changeset/mcp-honour-likec4-workspace-env.md
@@ -1,0 +1,5 @@
+---
+'likec4': patch
+---
+
+Fix `likec4 mcp` ignoring the `LIKEC4_WORKSPACE` environment variable. The command overrode the shared `path` default, so the workspace always fell back to the current directory and the server silently served an empty model. Behaviour now matches the documented help text and the other commands.
--- a/packages/likec4/src/cli/mcp/index.ts
+++ b/packages/likec4/src/cli/mcp/index.ts
@@ -15,7 +15,6 @@ const mcpCmd = <T>(yargs: Argv<T>) => {
         y
           .usage(`${k.bold('Usage:')} $0 mcp [path]`)
           .positional('path', path)
-          .default('path', '.', 'current directory')
           .option('stdio', {
             boolean: true,
             description: 'use stdio transport',


davydkov (approved):
Thanks!

PR #2949, opened 2026-05-11
fix: support bidirectional model relationships

ckeller42:
Fixes #2927.

## Summary

- Allow model and extend relationships to use `<->` and `-[kind]<->`.
- Preserve bidirectional semantics through styling and relationship extension matching.
- Include and exclude bidirectional model relationships from either endpoint in view predicates.
- Keep mixed parallel connections precise: a reverse incoming bidirectional match no longer pulls in a same-direction directed relation by accident.
- Update public docs and the LikeC4 DSL skill references.

## Examples / screenshots

All screenshots below were generated from minimal LikeC4 projects on this branch with `likec4 gen dot` and Graphviz.

### 1. Model relationship with `<->`

```likec4
model {
  component frontend 'Frontend'
  component api 'API'

  frontend <-> api 'Bidirectional sync'
}

views {
  view bidirectional_model {
    include frontend -> api
  }
}
```

Model relationship with <->

### 2. Incoming predicate from the declared target

```likec4
views {
  view incoming_api {
    include -> api
  }
}
```

The bidirectional `frontend <-> api` relationship is included when `api` is the incoming target.

include -> api includes the bidirectional relationship

### 3. Incoming predicate from the reverse endpoint

```likec4
views {
  view incoming_frontend {
    include -> frontend
  }
}
```

The same bidirectional relationship is also included when `frontend` is selected from the reverse endpoint.

include -> frontend includes the same bidirectional relationship

### 4. Mixed parallel regression guard

```likec4
model {
  component a 'Component A'
  component b 'Component B'

  a <-> b 'Bidirectional sync'
  a -> b 'Directed command'
}

views {
  view mixed_reverse_incoming {
    include -> a
  }
}
```

This must include only `Bidirectional sync`; the parallel one-way `Directed command` must not be pulled in by the reverse bidirectional match.

mixed parallel reverse incoming includes only the bidirectional relationship

## Testing

- `npx -y pnpm@11.5.1 generate`
- `npx -y pnpm@11.5.1 --filter @likec4/core test`
- `npx -y pnpm@11.5.1 --filter @likec4/language-server test`
- `npx -y pnpm@11.5.1 --filter @likec4/core typecheck`
- `npx -y pnpm@11.5.1 --filter @likec4/language-server typecheck`
- `npx -y pnpm@11.5.1 --filter @likec4/core test -- incoming-expr.spec.ts`
[...]

PR #3225, opened 2026-09-04
feat: support bidirectional relationships in deployment model

### LINKED ISSUE #3219: Bidirectional relationships not supported in deployment models
### Relates to

LikeC4 Language

### Version you are using

1.59.3

### Describe the issue

Even though this was probably only implemented for the logical model by design, I still classify it as a bug since essentially everything about deployment models follow the same rules/design as logical models.

_No response_

AndersHogqvist:
Extends the bidirectional relationship notation (`<->` and `-[kind]<->`) from PR #2949 to deployment relationships.

- like-c4.langium: DeploymentRelation reuses the shared RelationConnector fragment, so it accepts `<->` and `-[kind]<->` like logical relations.
- language-server: DeploymentModelParser and MergedSpecification propagate isBidirectional and default `tail` to render arrows on both ends.
- core: DeploymentRelationship/DeploymentRelationModel expose isBidirectional; the deployment-view incoming predicate can include and exclude bidirectional relations from either endpoint (findConnectionsFrom/ narrowToBidirectional); shared isBidirectionalRelation helper moved to compute-view/utils so element-view and deployment-view use the same logic.
- builder: NewDeploymentRelationProps no longer strips isBidirectional, so it can be set from the Builder API.
- generators: deployment and model relationship emitters share a relationConnector helper and emit `<->`/`-[kind]<->`; isBidirectional added to both zod schemas.
- docs/skill: documented bidirectional deployment relationships in apps/docs and the likec4-dsl skill references.
- tests: language-server parsing tests, a deployment-view compute spec for include/exclude symmetry, and a generator snapshot.
- added a changeset for @likec4/core, @likec4/language-server and @likec4/generators.

Fixes #3219

## Checklist


AndersHogqvist:
Just realized that there are things that need to be discussed here

PR #3226, opened 2026-09-04
feat(mcp): return links and sourceLocation from batch-read-elements

parse:
## Problem

`batch-read-elements` exists to cut round-trips, but returns neither `links` nor `sourceLocation`. `read-element` returns both.

So "which repo does each of these N elements live in", or "where is each declared", costs N `read-element` calls for data the batch tool is already shaped to return in one.

## Fix

Both fields are added to each element, from the code path `read-element` already uses: the same `computedModel(projectId)` and the shared `mkLocate` helper in `_common.ts`. `find-relationships` already emits per-item `sourceLocation` from a loop, so no new pattern is introduced.

The link mapping moves out of `read-element` into a shared `serializeLinks` in `_common.ts`, now used by `read-element`, `read-deployment` and `batch-read-elements`. Output of the first two is unchanged.

Shapes match `read-element` exactly, so callers handle one shape for every tool:

- `sourceLocation.range.start.line` stays 0-based. Normalizing it here would break existing `read-element` consumers.
- An element with no links returns `[]`, never an omitted field.

Additive only, no new tool. Response size is not a concern: `includedInViews` already dominates the payload, and links plus a source location are a few bytes per element.

## Changes

- **`@likec4/mcp`**: the two fields on `batch-read-elements`, shared `serializeLinks` / `linksSchema` in `_common.ts`, tool description, server instructions and README updated to list them
- **`apps/docs`**: `batch-read-elements` row in the MCP tool table

## Tests

Batch output for a given element is asserted equal to `read-element` output for both fields, so the two cannot drift. Plus link cases: an element with two links, one with a relative link, one with none, and a mix in a single batch call.

## Checklist

### DIFF
--- a/.changeset/mcp-batch-read-elements-links.md
+++ b/.changeset/mcp-batch-read-elements-links.md
@@ -1,0 +1,5 @@
+---
+'@likec4/mcp': patch
+---
+
+`batch-read-elements` now returns `links` and `sourceLocation` for each element, matching `read-element`. Reading those for many elements no longer needs one call per element.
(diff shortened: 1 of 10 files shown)


davydkov (approved):
Hey @parse, thanks for contribution!

PR #3228, opened 2026-09-05
fix: honor workspace exclusions during filesystem discovery

MandarinOptimizer:
Workspace exclusions are currently checked when assigning files to projects, but the Node filesystem provider does not use them while crawling for project configuration and model files. Consequently, an excluded generated/runtime directory is still traversed during startup and project reloads.

This change passes the shared services to the filesystem provider and consults the current workspace exclusion matcher before descending into directories, before scanning an excluded root, and when selecting individual files. The existing built-in directory exclusions and supported directory-symlink discovery remain intact. Project-specific exclusions and Git ignore handling are unchanged.

Validation:
- Six synthetic filesystem regression tests fail on the original implementation and pass with the fix. Directory-read spies distinguish traversal pruning from filtering results afterward.
- The language-server suite passes: 823 passed, 3 skipped, 2 todo across 66 files.
- `pnpm exec tsc -b packages/language-server` passes.
- Scoped oxlint and dprint checks pass.

The tests use small temporary fixtures only; no real workspace files, logs, data, or screenshots are included. This fixes exclusion-aware discovery; it does not change language-client crash/restart policy or introduce automatic `.gitignore` support.

### DIFF
--- a/.changeset/tidy-workspace-discovery.md
+++ b/.changeset/tidy-workspace-discovery.md
@@ -1,0 +1,6 @@
+---
+"@likec4/language-server": patch
+"likec4-vscode": patch
+---
+
+Honor editor workspace exclusions before traversing folders during project and model discovery, avoiding unnecessary scans of excluded generated files and runtime storage.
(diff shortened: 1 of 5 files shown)


davydkov (approved):
Thanks for the fix!

PR #3238, opened 2026-09-09
fix: respect dynamic view variant on initial render

### LINKED ISSUE #3202: "variant sequence" attribute of dynamic views ignored in HTML export
### Relates to

CLI

### Version you are using

1.59.2

### Describe the issue

Steps to reproduce:

1. Create a dynamic view.
2. Set the "variant" attribute to "sequence".
3. The attribute is properly applied in VS Code preview.
4. Export project to HTML (npx -y likec4@latest build -o "./html" --output-single-file).
5. Open the view.
6. The visualization defaults to the "diagram" type (not sequence).

_No response_

cry999:
## Summary

Fix the initial rendering of dynamic views so that their configured `variant` is respected when the URL does not explicitly specify a `dynamic` parameter.

Previously, the search parameter parser always defaulted `dynamic` to `diagram`. This caused the SPA to override `variant sequence` defined in the LikeC4 source.

The URL parameter still takes precedence when explicitly provided, for example `?dynamic=diagram`.

Fixes #3202

## Changes

- Leave the parsed `dynamic` search parameter undefined when it is absent or invalid
- Allow the diagram state to select the variant configured on the view during initial rendering
- Preserve explicit URL overrides
- Add an E2E fixture and regression coverage for:
  - `variant sequence` without a `dynamic` URL parameter
  - `?dynamic=diagram` overriding `variant sequence`

## Verification

- [x] SPA typecheck
- [x] SPA unit tests
- [x] Search parameter E2E tests
- [x] Manual single-file build verification
- [x] Changeset added

## Notes

This change only affects initial rendering. The existing navigation behavior remains unchanged: the display variant selected on the previous page takes precedence over the `variant` configured on the destination view.


davydkov (approved):
@cry999  Thanks for the fix!

PR #3244, opened 2026-09-10
fix(generators): emit valid Mermaid for views with a group

### LINKED ISSUE #3243: gen mmd emits invalid Mermaid for views with a group: node ids get sliced by the group id's length
### Describe the bug

`likec4 gen mmd` produces Mermaid that no renderer accepts for any view containing a view-level `group` block. Two separate problems in the same output:

1. The generated subgraph id is `@gr1`. Mermaid's lexer rejects a leading `@` — it is the sigil for the shape-data syntax (`node@{ … }`) — so the diagram fails to parse at the `subgraph` line.
2. Each grouped element's id is emitted as the element id with `groupId.length` characters removed from the front, as though the element id were prefixed by the group id. `@gr1.` is 5 characters, so `portal` comes out as `L` and `auth` — being shorter than 5 — comes out as the **empty string**.

Both symptoms disappear as soon as the `group` block is removed; every ungrouped view in the same project generates fine.

### Reproduction

`likec4.config.json`:

```json
{ "$schema": "https://likec4.dev/schemas/config.json", "name": "repro" }
```

`model.c4`:

```likec4
specification {
  element system
}

model {
  app = system 'App'
  auth = system 'Auth'
  portal = system 'Portal'

  app -> auth 'signs in'
  app -> portal 'authz'
}

views {
  view index {
    include app
    group 'Infra' {
      include auth, portal
    }
  }
}
```

Then:

```sh
docker run --rm -v "$PWD:/data" ghcr.io/likec4/likec4:1.59.3 gen mmd -o /data/out /data
```

### Actual output

```mermaid
---
title: "Landscape view"
---
graph TB
  App@{ shape: rectangle, label: "App" }
  subgraph @gr1["`Infra`"]
    @gr1.@{ shape: rectangle, label: "Auth" }
    @gr1.L@{ shape: rectangle, label: "Portal" }
  end
  App -. "`signs in`" .-> @gr1.
  App -. "`authz`" .-> @gr1.L
```

Note `@gr1.` used as a node id on lines 7 and 10 — `auth` has lost its identifier entirely, and `portal` has lost its first five characters.

### Expected output

Ids that are valid Mermaid and that still distinguish the two elements, e.g. `_gr1.Auth` / `_gr1.Portal` under `subgraph _gr1`.

### Rendering

Kroki (Mermaid 11.16.0):

```
Error 400: SyntaxError: Lexical error on line 3. Unrecognized text.
... "App" }  subgraph @gr1["`Infra`"]    
----------------------^
```

### Environment

- LikeC4 1.59.3 (`ghcr.io/likec4/likec4:1.59.3`)
- `likec4 validate` and `likec4 format --check` both pass on the source; only the Mermaid generator is affected. The `dot`, `d2` and `plantuml` generators were not checked.

### Impact

A project that uses `group` for a real architectural boundary cannot export those views to Mermaid at all. Our workaround was to drop the view-level groups and model the boundary as a parent element instead, which changes how neighbouring elements are derived in every other view — a fairly invasive change to work around an exporter bug.

humb1t:
Fixes #3243

## Checklist

## The problem

`likec4 gen mmd` emits Mermaid that no renderer accepts for any view containing a view-level `group`. Two defects, one code path:

1. The subgraph id is `@gr1`. Mermaid reserves a leading `@` for the `node@{ ... }` shape syntax, so its lexer fails on the `subgraph` line.
2. Each grouped element's id is emitted with `groupId.length` characters removed from the front. `@gr1.` is 5 characters, so `portal` comes out as `L`, and `auth` — being shorter than 5 — comes out as the **empty string**, which is then used as a node id twice.

## Root cause

`packages/generators/src/mmd/generate-mmd.ts:16`

```ts
const nodeName = (node: Node): string => {
  return fqnName(node.parent ? node.id.slice(node.parent.length + 1) : node.id)
}
```

The slice assumes `node.parent` is a prefix of `node.id`, which holds for genuinely nested elements (`cloud` / `cloud.backend`). It does not hold for view groups: the group id is synthetic (`@gr${memory.groups.length + 1}`, `packages/core/src/compute-view/element-view/memory/memory.ts:113`) and members keep their own top-level ids, so the slice cuts into the member's real id.

## The change

- Only strip the parent prefix when the child id actually carries it.
- Sanitise generated ids to word characters, so `@` (and anything else outside `\w`) can never reach the output.

Existing snapshots for nested views are unchanged, so ordinary nesting keeps producing the same ids.

## Verification

Repro from the issue, generated with `pnpm cli gen mmd` from this branch:

```mermaid
---
title: "Landscape view"
---
graph TB
  App@{ shape: rectangle, label: "App" }
  subgraph _gr1["`Infra`"]
    _gr1.Auth@{ shape: rectangle, label: "Auth" }
    _gr1.Portal@{ shape: rectangle, label: "Portal" }
  end
  App -. "`signs in`" .-> _gr1.Auth
  App -. "`authz`" .-> _gr1.Portal
```

- Before the change, POSTing the generated file to a local Kroki `/mermaid/svg` returns `400 SyntaxError: Lexical error on line 3. Unrecognized text.` pointing at the `@` in `subgraph @gr1`. After, it returns `200` and the SVG contains all four labels.
- `pnpm test` — 294 files, 3008 passed.
- `pnpm typecheck` — 26/26 tasks successful.
- The new test fails on `main` and passes with the fix.

## Note, not fixed here

`packages/generators/src/d2/generate-d2.ts:12` has the same `node.id.slice(node.parent.length + 1)`, so the D2 generator likely truncates grouped element ids too. I left it alone to keep this PR scoped to the reported issue — happy to extend it if you'd prefer one PR.

---

Disclosure: this change was written with AI assistance. The repro, the fix and the verification above were all executed and confirmed against this branch.

### DIFF
--- a/.changeset/fix-mermaid-view-groups.md
+++ b/.changeset/fix-mermaid-view-groups.md
@@ -1,0 +1,5 @@
+---
+'@likec4/generators': patch
+---
+
+Fix Mermaid export producing invalid output for views with a `group`. The subgraph id no longer starts with `@`, which Mermaid reserves for the `node@{ ... }` shape syntax, and grouped elements keep their own ids instead of being truncated by the length of the generated group id (which left short ids, such as `auth`, with an empty id).
--- a/packages/generators/src/__mocks__/data.ts
+++ b/packages/generators/src/__mocks__/data.ts
@@ -524,3 +524,60 @@ Label`.trimStart(),
   viewOf: null,
   width: 600,
 } as any
+
+export const fakeComputedViewWithGroups: ComputedView = {
+  id: 'index',
+  title: 'Landscape view',
+  nodes: [
+    {
+      id: 'app',
+      parent: null,
+      title: 'App',
+      children: [],
+      color: 'primary',
+      shape: 'rectangle',
+    },
+    {
+      id: '@gr1',
+      parent: null,
+      kind: '@group',
+      title: 'Infra',
+      children: ['auth', 'portal'],
+      color: 'primary',
+      shape: 'rectangle',
+    },
+    {
+      id: 'auth',
+      parent: '@gr1',
+      title: 'Auth',
+      children: [],
+      color: 'primary',
+      shape: 'rectangle',
+    },
+    {
+      id: 'portal',
+      parent: '@gr1',
+      title: 'Portal',
+      children: [],
+      color: 'primary',
+      shape: 'rectangle',
+    },
+  ],
+  edges: [
+    {
+      id: 'app:auth',
+      source: 'app',
+      target: 'auth',
+      label: 'signs in',
+    },
+    {
+      id: 'app:portal',
+      source: 'app',
+      target: 'portal',
+      label: 'authz',
+    },
+  ],
+  autoLayout: { direction: 'TB' },
+  rules: [],
+  viewOf: null,
+} as any
--- a/packages/generators/src/mmd/__snapshots__/generate-mmd.spec.ts.snap
+++ b/packages/generators/src/mmd/__snapshots__/generate-mmd.spec.ts.snap
@@ -83,3 +83,18 @@ graph TB
   Client -. "\`opens\`" .-> SystemFrontend
 "
 `;
+
+exports[`generate mermaid - view with group 1`] = `
+"---
+title: "Landscape view"
+---
+graph TB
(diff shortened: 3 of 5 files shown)


davydkov (approved):
Hey @humb1t,  thanks for the fix!

PR #3266, opened 2026-09-19
fix(core): apply relationship kind default title to dynamic view steps

### LINKED ISSUE #3229: Relationship kind's title default not applied in dynamic view steps (unlike technology, which works correctly)
### Relates to

LikeC4 Language

### Version you are using

1.59.3

### Describe the issue

When a relationship kind defines a default `title` in `specification { }`, that default is not applied to relationships used as dynamic view steps — unless the relationship also explicitly repeats the same title inline. The kind's technology default, by contrast, is applied correctly in the same scenario.

```
specification {
    element container

    relationship discover {
        title ":8761"
        technology "HTTP"
    }
}

model {
    gatewayService1 = container 'Gateway Service 1'
    gatewayService2 = container 'Gateway Service 2'
    discoveryServer = container 'Discovery Server 1'
}

views {
    dynamic view test{
        // Step 1: explicit inline title — renders correctly
        gatewayService1 -[discover]-> discoveryServer ':8761'

        // Step 2: relies ONLY on the kind's default title — title does not render
        gatewayService2 -[discover]-> discoveryServer
    }
}
```

Image

_No response_

chiliec:
Closes #3229

## What

When a relationship kind defines a default `title` in `specification { }`, a dynamic view step using that kind (e.g. `gatewayService2 -[discover]-> discoveryServer`) did not render it, while the kind's `technology` default was applied.

## Fix

In `packages/core/src/compute-view/dynamic-view/compute.ts`, the computed step title now falls back to the kind specification after the step's own title and the title derived from model relationships:

```ts
title: stepTitle ?? title ?? kindSpec?.title,
```

Precedence is unchanged otherwise: explicit step title > model relationship title > kind default.

## Tests

Added a `Title inheritance` block to `packages/core/src/compute-view/dynamic-view/__test__/step-fields.spec.ts`:
- kind default title is applied when the step has a kind but no title (fails on `main`: `label: null`, passes with the fix)
- explicit step title still wins over the kind default
- model relationship title still wins over the kind default

## Validation

- `pnpm exec vitest run src/compute-view` in `packages/core`: 68 files, 554 passed | 1 expected fail | 1 todo
- `tsc -p tsconfig.json --noEmit` in `packages/core`: clean
- `dprint check` on changed files: clean

## Checklist

- [x] I've verified `pnpm typecheck` and `pnpm test` (for `@likec4/core`).


davydkov (approved):
Hey @chiliec thank you for the fix!

PR #3267, opened 2026-09-19
fix(language-server): allow empty title and technology in view customization

### LINKED ISSUE #3197: Hide relationship title and technology; hide element title
### Relates to

Diagrams (React)

### Version you are using

1.59.2

### Describe the issue

Hiding relationship title and technology and hiding element title doesn't work in views when using `with` per https://github.com/likec4/likec4/issues/910.

Passing in an empty string `''` has no effect on relationship titles or technologies, and passing in an empty string has no effect on elements. Hiding element descriptions does however work. Providing a non-empty string works as expected, ie. the title or technology is replaced by the string.

```
specification {
  element actor
  element system
}

model {
  customer = actor 'Customer' {
    description 'The regular customer of the system'
  }

  saas = system 'Our SaaS'
  customer -> saas 'enjoys our product' {
    technology 'HTTPS'
  }
}

views {

  view index {
    title 'Landscape view'

    include customer with {
      title ''
      description ''
    }

    include saas

    include * -> * with {
      title ''
      technology ''
    }
  }
}
```

https://playground.likec4.dev/share/YeTHq6K4IF/

chiliec:
Closes #3197

## What

Inside `with { ... }` (both element and relationship customization), `title ''` and `technology ''` had no effect: the parser dropped the parsed value with a truthiness check (`if (parsed)`), so an empty string was treated the same as "not set". `description ''` already worked because it goes through `parseMarkdownOrString`, which returns `{ txt: '' }`.

## Fix

`packages/language-server/src/model/parser/FqnRefParser.ts`: skip only `undefined` (`isDefined(parsed)`) in `parseFqnExprWith` and `parseCustomRelationProperties`, so an explicit empty string is passed through. The diagram side already hides empty values (`isTruthy(label)` / `isTruthy(technology)` in `EdgeLabel`, `ElementData`), and the core compute layer already handles `label: ''` (existing test "set edge title to empty string").

## Tests

`packages/language-server/src/model/__tests__/issue-3197.spec.ts` builds the model from the issue and asserts `label: ''` / `technology: ''` on the edge and `title: ''` / `technology: ''` on the node. Both tests fail on `main` and pass with the fix.

```
pnpm --filter @likec4/language-server exec vitest run src/model/__tests__/issue-3197.spec.ts
  Tests  2 passed (2)       # with fix
  Tests  2 failed (2)       # on main
pnpm --filter @likec4/language-server test
  Test Files  65 passed, Tests  779 passed | 3 skipped | 2 todo
```

Two tests in `fqn-ref-expr.spec.ts` hit the 30s timeout during the full run on my 8 GB box; the file passes in isolation (8/8), unrelated to this change. I could not run `tsc -b` for the package locally (OOM on this machine).


davydkov (approved):
Hey @chiliec thanks for the fix!

